PR #6, opened 2018-05-02
update yml file 

martham93:
adding packages: nltk, tweepy 

@lwasser

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -40,3 +40,6 @@ dependencies:
       - tqdm
       - kiwisolver
       - git+https://github.com/geopandas/geopandas.git
+      #text mining
+      - tweepy
+      - nltk 


lwasser:
Thank you @martham93 !!!

lwasser:
hey @martham93  just a quick question. here you have tweepy and nltk installed via pip. is there a reason for that vs conda-forge? we try to install as much as possible from conda-forge but if pip is better for some reason please let me know. otherwise i'll move these to conda-forge

martham93:
hello @lwasser! No reason for using pip, I just forgot that we try to only use conda-forge.

lwasser:
Ok fantastic. Thank you @martham93 . i will accept this and then update it for pip. I appreciate the quick response :)

PR #18, opened 2018-10-12
Support testing on Travis CI + Codecov

mbjoseph:
Ok @lwasser! This PR adds support for testing the earthpy package on Travis CI, and uploads code coverage reports onto Codecov upon successful builds. 

Tests are contained in the `earthpy/tests/` directory, and for starters I've added just one test to get going. To run these tests locally, you can run `pytest` from the command line (the `environment.yml` file has been updated to install the pytest package, so you may need to update your environment). I spent a fair bit of time comparing testing frameworks, and pytest seems to be one of the better ones (simple to use but powerful when you need it to be).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,14 +23,10 @@ Then import it into python.
 
 ## Contributors
 
-This package was originally developed by Chris Holdgraf.
-Contributors:
-
-- Carson Farmer
-
-Contributing Breakers:
-
+- Chris Holdgraf
 - Leah Wasser
+- Carson Farmer
+- Max Joseph
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,4 @@ Contributing Breakers:
 ## Testing
 
 This package uses [pytest](https://pytest.org/) for tests.
-To run tests locally, you can exectute the command `pytest`.
-Tests will also be run on Travis CI, and upon a successful build test coverage
-information will be uploaded to Codecov.
+To run tests locally, execute the command `pytest` from the command line.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -29,3 +31,8 @@ Contributors:
 Contributing Breakers:
```
lwasser: @mbjoseph  will you kindly add yourself as a contributor here? 
Perhaps we should just have a contributor list and include Chris H and me :) i'm trying not to be a breaker these days!

environment.yml
```diff
@@ -3,10 +3,13 @@ channels:
   - conda-forge
```
lwasser: i'm curious @mbjoseph does this envt need to be in this repo? i have a full version of the envt that i maintain elsewhere is why i'm asking but that envt doesn't have pytest in it as the students don't need it. thoughts?

README.md
mbjoseph: Sure thing! I agree it's probably better not to split hairs between contributors and breakers :)

mbjoseph: Fixed with https://github.com/earthlab/earthpy/pull/18/commits/c4f9cf5059b3c630c8470d94601ac0f4b8858ab8

environment.yml
mbjoseph: Good question.... I think you would want to have some kind of file that specifies the requirements for the earthpy package in this repo. This doesn't necessarily need to be identical to the environment that you use for teaching though. 

I would guess that the minimal set of requirements for earthpy is a subset of all of the packages you would want to use in your class, right? e.g., you want to use Jupyter Lab to teach with, but it's not necessary for earthpy to run. So, I wonder whether there is a way to have a requirements file for earthpy that lives on its GitHub repo, and then a separate environment.yml file for your class?

mbjoseph: @lwasser maybe we can open another issue to figure out where then evironment.yml file should live? Seems to be a separate problem than adding Travis CI + Codecov to the repo. We can change the environment.yml file later on and adjust Travis CI and Codecov to match our solution later on once we reach a decision.

Issue #19, opened 2018-10-15
Finding a home for earthpy's dependencies


mbjoseph:
Currently this repo contains an environment.yml file that is very similar (but not identical) to the [`environment.yml` file](https://github.com/earthlab/earth-analytics-python-env/blob/master/environment.yml) for the earth analytics class conda environment.

But, this package has dependencies that aren't needed for the class, and the class has dependencies that aren't needed for this package. One solution to this is to deal with the environment for this package separately from the environment for the class (but still make sure that students using the class environment can use this package). 

From what I can gather from https://python-packaging.readthedocs.io/en/latest/dependencies.html, the convention is to use `install_requires` in `setup.py` (https://github.com/earthlab/earthpy/blob/master/setup.py#L33-L34) to specify the minimal set of requirements for users to install and use the package. But, it also seems like we could use a `requirements.txt` file or an `environment.yml` file. 

**So my question is:** Which of the three options (`install_requires`, `requirements.txt`, and `environment.yml`) is better in this case for ensuring that this package complies with typical practice, and that the package works with the earth analytics class conda environment?

betatim:
As a user of this package I would expect `pip install earthpy` to take care of installing all the dependencies (except for non Python packages). This means the dependencies should be listed in `install_requires`.

For dependencies that can't be installed by `pip` (non Python packages) there are usually instructions in the `README` and a check in `setup.py` to try and determine if those dependencies were installed or not and output a message telling users that they need to fix up something else first.

lwasser:
@betatim can we use conda-forge rather than pip for dependencies? can that be specified? i am asking because of the known errors with pip and spatial libraries.

mbjoseph:
We are going to use conda :snake:

PR #25, opened 2018-10-18
[WIP] Use setup.py to specify dependencies

mbjoseph:
This work in progress PR uses setup.py instead of the earth-analytics-python conda environment to manage earthpy's dependencies and relates to https://github.com/earthlab/earthpy/issues/19. I have tested this branch locally by installing earthpy in a fresh virtual environment and also by installing earthpy from source in the earth-analytics-python environment. Both seem to work well, but in the absence of real unit tests, it's hard to know whether there are potential issues that aren't being flagged as errors in the installation process. 

Before merging this, I think we need to have real meaningful unit tests that will tell us whether there are going to be problems. Maybe @lwasser, @joemcglinchy, @jlpalomino, and @lherwehe and I can discuss a strategy in our Monday meeting for testing the code that we already have in earthpy. 

Once we have real unit tests, we will know whether this approach (vs. explicitly using conda-forge) for dependency management is going to be problematic.

- [ ] Check to make sure that unit tests pass (once they exist)


setup.py
```diff
@@ -9,6 +8,18 @@
...
+requirements = [
+    'download',
+    'geopandas',
+    'matplotlib',
+    'numpy',
+    'pandas',
+    'rasterio',
+    'shapely',
+    'scikit-image',
+    'tqdm'
```
lwasser: we might be able to drop tqdm although the message is pretty nice  when you download. i will think about this.  Also can you kindly update your branch from master as we are on version .3 now!! :)

setup.py
```diff
@@ -30,8 +41,17 @@ def configuration(parent_package='', top_path=None):
           maintainer_email=MAINTAINER_EMAIL,
           description=DESCRIPTION,
           version=VERSION,
-          install_requires=['tqdm', 'pandas', 'numpy', 'geopandas',
-                            'matplotlib', 'rasterio', 'download'],
+          install_requires=requirements,
```
lwasser: +1

.travis.yml
```diff
@@ -6,20 +6,7 @@ python:
...
-  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
-  - bash miniconda.sh -b -p $HOME/miniconda
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
-  - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
-  # Useful for debugging any issues with conda
-  - conda info -a
-  - conda env create -f environment.yml
-  - source activate earth-analytics-python
```
lwasser: The thing that i'm worried about here is that our envt is specific to this library. Many of the dependencies - like GDAL won't install if people start using pip. conda-forge is the main install repo. i just don't understand how this will work without an explicit envt!

mbjoseph: I think I am following, and I totally hear you that conda-forge would be the go-to place from which to install earthpy's dependencies. Having unit tests will help us to figure out whether we get different answers if we install using pip vs. conda/conda-forge.

As far as multiple environments - I think the goal here is to capture earthpy's dependencies in this repo, and capture your earth analytics class dependencies elsewhere. That way, we check two boxes:

1. Students can still install your conda environment, then install this package, and have most or all of earthpy's dependencies installed already. 

2. Users of this package do not need to install the full earth-analytics-python conda environment (e.g., if you only wanted to use earthpy, and not install all of the other packages that get used in the course).

betatim: The way I understand Max's comment: have earthpy (this package) specify its dependencies in `setup.py` plus instructions for manual steps (often this is left to the reader "Make sure you have GDAL>=vX.Y, how you get it is up to you") for things that can't be pip installed (e.g. GDAL which is needed by rasterio).

Then we have the earth-analytics-python which is a ready made environment that has an opinion on how you should install stuff (conda from conda-forge). In practice not much would have to change in `earthlab/earth-analytics-python-env` as it `pip` installs earthpy which if all the dependencies are already satisfied will just work.

Overall I think Max's plan sounds good.

PR #44, opened 2018-10-23
switched parameters for normalized_diff function

joemcglinchy:


lwasser:
Thank you @joemcglinchy ... my students are actively working on an assignment this week. So i think for now i will hold off on merging this change...

I'm debating whether to just tell them i've made the change next week? OR to let them finish out the last homework (next week) and then make the change. I'm inclined just to wait a few weeks to implement this. Are you ok with that?

PR #45, opened 2018-10-23
Tif gen for testing

joemcglinchy:
added a couple of helper functions to test_spatial.py to support:
- creating 3D arrays where the pixel value for each band is the number of that band createArray
- writing a 3D array to a GeoTIFF file on disk

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -40,7 +40,7 @@ def extent_to_json(ext_obj):
 
 # calculate normalized difference between two arrays
 # both arrays must be of the same size
-def normalized_diff(b1, b2):
+def normalized_diff(b2, b1):
     """Take two numpy arrays and calculate the normalized difference
     Math will be calculated (b2-b1) / (b2+b1).
 
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -40,7 +40,7 @@ def extent_to_json(ext_obj):
 
 # calculate normalized difference between two arrays
 # both arrays must be of the same size
-def normalized_diff(b2, b1):
+def normalized_diff(b1, b2):
     """Take two numpy arrays and calculate the normalized difference
     Math will be calculated (b2-b1) / (b2+b1).
 
--- a/earthpy/tests/test_spatial.py
+++ b/earthpy/tests/test_spatial.py
@@ -1,5 +1,88 @@
 from earthpy import spatial
+from osgeo import gdal
+from osgeo import osr
+import rasterio as rio
+import numpy as np
 
+# a helper function to generate a 3D array of data
+def createArray(arr_shape):
+    """ createArray(arr_shape) creates a numpy ndarray with the shape defined by the tuple arr_shape
+        arr_shape should be:
+        (rows, cols, [channels])
+        the values for each channel will be the 1-based number of that channel.
+    """
+    
+    dummy_arr = np.ones(arr_shape)
+    
+    if len(arr_shape) == 2:
+        
+        # there is only one channel, so that one channel is all 1's
+        dummy_arr = np.expand_dims(dummy_arr,-1)
+    
+    else:
+        # multiply each channel by the channel number (1-based)
+        for i in range(arr_shape[-1]):
+            dummy_arr[:,:,i] *= i+1
+    
+    return dummy_arr
+
+# a helper function to write a 3D array of data to disk as a GeoTIFF 
+# the result will have no spatial info or relevance
+def dummy_tif_writer(arr, dst_filename):
+    """
+        dummy_tif_writer(arr_dst_filename) writes a tif file specified by dst_filename using the array specified
+        by arr. The shape of arr should be 3 dimensional such that the last dimension is the number of bands.
+        this type of array can be created with the createArray(arr_shape) function.
+    """
+
+    # make sure the array is 3 dimensional, assuming last dimension is number of bands
+    # this is assured by using createArray to generate the data
+    assert len(arr.shape) == 3
+    
+    try:
+        # define a dummy geotransform
+        geotrans = (0,1,0,0,0,-1)
+
+        # get the dimensions of the array
+        n_rows, n_cols, n_channels = arr.shape
+        n_rows, n_cols, n_channels
+
+        x_pixels = n_cols  # number of pixels in x
+        y_pixels = n_rows  # number of pixels in y
+
+        # set the driver
+        driver = gdal.GetDriverByName('GTiff')
+
+        # create the file
+        dataset = driver.Create(dst_filename, x_pixels, y_pixels, int(n_channels), gdal.GDT_Byte)
+
+        # write the bands
+        for i in range(n_channels):
+            dataset.GetRasterBand(i+1).WriteArray(dummy[:,:,i])
+
+        # define the Unknown projection
+        srs = osr.SpatialReference()            
+        srs.ImportFromEPSG(0)
+        proj = srs.ExportToWkt() 
+
+        # set the geotransform
+        dataset.SetGeoTransform(geotrans)
+
+        # set the projection
+        dataset.SetProjection(proj)
+
+        # flush the cache
+        dataset.FlushCache()
+
+        # remove the dataset from memory
+        dataset=None
+    
+        return (0, dst_filename)
+    
+    except Exception as e:
+        
+        return(e, None)
+    
 
 def test_dummy():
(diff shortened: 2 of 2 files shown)


earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
 from earthpy import spatial
+from osgeo import gdal
+from osgeo import osr
+import rasterio as rio
+import numpy as np
 
+# a helper function to generate a 3D array of data
+def createArray(arr_shape):
```
mbjoseph: @joemcglinchy this looks great! I wonder whether we can remove the `createArray` function, since I think we will want to be able to make more types of dummy arrays than what this currently implements, and we can do that pretty easily using `np.ones()` one-liners + the assertion that you wrote in `dummyTifWriter`. For example, if we need to make arrays that have weird bounds to test byte scaling, or arrays with negative values we can do that with `np.random.uniform(..., size=(rows, cols, channels))` or `np.ones((rows, cols, channels)) * -1`.

earthpy/spatial.py
```diff
@@ -40,7 +40,7 @@ def extent_to_json(ext_obj):
 
 # calculate normalized difference between two arrays
 # both arrays must be of the same size
-def normalized_diff(b1, b2):
+def normalized_diff(b2, b1):
```
lwasser: @joemcglinchy can you kindly remove this fix in this PR so it's distinct from the tif generator? I'd like to hold off on implementing this for my students sake! but i want to implement it. just not in this PR.

earthpy/tests/test_spatial.py
lwasser: i'm sorry to muck up the convo here with my syntax stickling ... but can we please follow PEP8 guidelines

https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

Rather than camel case i prefer all `lower_case_with_underscore` naming conventions. Please yell if this is not in line with your thoughts on this package.

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+    else:
+        # multiply each channel by the channel number (1-based)
+        for i in range(arr_shape[-1]):
+            dummy_arr[:,:,i] *= i+1
+    
+    return dummy_arr
+
+# a helper function to write a 3D array of data to disk as a GeoTIFF 
+# the result will have no spatial info or relevance
+def dummyTifWriter(arr, dst_filename):
```
lwasser: Let's also follow docstring conventions. 

```
    """Writes a tif file to a specified location using an array.

    Parameters
    ----------
    arr : array 
        More here on the array.
    dst_filename: path to file...
        More here on the path
```
Please see rasterio docs for examples!
Also can we please use destfile as the argument? i've been using it elsewhere.

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+
+        # remove the dataset from memory
+        dataset=None
+    
+        return (0, dst_filename)
+    
+    except Exception as e:
+        
+        return(e, None)
+
```
mbjoseph: @joemcglinchy we should add tests to make sure this dummy tif writer works as we expect. This will fix the build which is currently broken because it reduces test coverage.

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+    # this is assured by using createArray to generate the data
+    assert len(arr.shape) == 3
+    
+    try:
+        # define a dummy geotransform
+        geotrans = (0,1,0,0,0,-1)
+
+        # get the dimensions of the array
+        n_rows, n_cols, n_channels = arr.shape
+        n_rows, n_cols, n_channels
```
mbjoseph: Can this line be removed?

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+        this type of array can be created with the createArray(arr_shape) function.
+    """
+
+    # make sure the array is 3 dimensional, assuming last dimension is number of bands
+    # this is assured by using createArray to generate the data
+    assert len(arr.shape) == 3
+    
+    try:
+        # define a dummy geotransform
+        geotrans = (0,1,0,0,0,-1)
```
mbjoseph: Probably more efficient to use an expressive name here instead: 

```python
dummy_geotransform = (0, 1, 0, 0, 0, -1)
```

That way you don't need to have a comment explaining what is being defined!

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+
+        # get the dimensions of the array
+        n_rows, n_cols, n_channels = arr.shape
+        n_rows, n_cols, n_channels
+
+        x_pixels = n_cols  # number of pixels in x
+        y_pixels = n_rows  # number of pixels in y
```
mbjoseph: Can we delete these lines and just use `n_cols` and `n_rows` to simplify?

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+
+        # set the driver
```
mbjoseph: This comment can be removed IMO

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+
+        # set the geotransform
+        dataset.SetGeoTransform(geotrans)
+
+        # set the projection
+        dataset.SetProjection(proj)
+
+        # flush the cache
+        dataset.FlushCache()
+
```
mbjoseph: Comments like `# set the geotransform`, `# set the projection` and `# flush the cache` can be removed - it's clear from the code what's happening!

earthpy/tests/test_spatial.py
```diff
@@ -1,5 +1,88 @@
...
+        # remove the dataset from memory
+        dataset=None
+
```
mbjoseph: This comment is good because it explains **why** this is being done, but I think you want to have spaces around the assignment operator.

earthpy/tests/test_spatial.py
mbjoseph: Some things to test: 

- Does it actually make a GeoTIFF?
- Does it raise errors where you want it to?

earthpy/tests/test_spatial.py
joemcglinchy: done!

earthpy/tests/test_spatial.py
```diff
@@ -1,6 +1,79 @@
...
+    destfile = 'dummy.tif'
+    arr = np.ones((5,5,1))
+    code,fi = dummy_tif_writer(arr, destfile)
+    
+    assert code == 0
+    assert os.path.exists(destfile) is True
+    
+    # Clean up the file
+    if os.path.exists(destfile):
+        os.remove(destfile)
```
lwasser: @mbjoseph it looks like there are tests here. Are there other items you'd like to see in this PR or can we merge it?

earthpy/tests/test_spatial.py
lwasser: oops just noticed this actually is still using camelCase rather than snake_case. camel being reserved for class names only. can you kindly fix @joemcglinchy

joemcglinchy:
The latest commit should have this function removed

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

earthpy/tests/test_spatial.py
mbjoseph: This test looks good to make sure it works how we expect it to, but I'd love to see one more test to make sure it *doesn't* work how we expect it to not work, i.e., a test that hits the `except` part of the `try...except...` logic of the function. @joemcglinchy can you add a test case that returns the exception that you implemented at the end of `dummy_tif_writer` (or remove the exception if it's not needed)?

lwasser:
@joemcglinchy  this is currently throwing an assertion error -- see below or [here](https://travis-ci.org/earthlab/earthpy/builds/452577636#L1204)
```
def test_dummy_tif_writer():
        """ Testing dummy_tif_writer."""
    
        destfile = 'dummy.tif'
        arr = np.ones((5,5,1))
        code,fi = dummy_tif_writer(arr, destfile)
    
>       assert code == 0
E       assert NameError("name 'dst_filename' is not defined",) == 0
```

I fixed the other error but can you kindly look into this error and also update your fork from earthpy/master as well?

Issue #39, opened 2018-10-22
Write unit test for bounding box JSON generation


mbjoseph:
https://github.com/earthlab/earthpy/blob/d2c1e31ebb6915ce281ea4b5af524511272abb70/earthpy/spatial.py#L16

First test for an example.

lwasser:
I believ that this has been IMPLEMENTED and can be closed! closing but please do re-open if it should remain open @mbjoseph

Issue #34, opened 2018-10-22
Write Unit Test & Update Asserts for crop_image


lwasser:

mbjoseph:
Note to whoever writes these tests, please consider the edge case of cropping using an extent that is just a point (with zero area), e.g., something like the result of 

```python
earthpy.spatial.extent_to_json([0, 0, 0, 0]) # [xmin, ymin, xmax, ymax]
```

For completeness, there is also the edge case of when the extent is a line, e.g., 

```python
earthpy.spatial.extent_to_json([0, 0, 1, 0])
```

mbjoseph:
This issue is a duplicate of  #70

PR #46, opened 2018-10-23
Adding unit tests for extent_to_json

mbjoseph:
This adds unit tests for `extent_to_json` to verify that it can create
extent JSON using both list and GeoDataFrame inputs, and that these
extent outputs are the same regardless of input type. I have also fixed
an error in the implementation of the list-based method (the elements of
the list needed to be unpacked with a `*` in order to act as positional
arguments to the `box` function.

The unit tests also include a check to ensure that a `ValueError` is
raised when we provide input that is neither a list nor a GeoDataFrame
(notice the use of the `pytest.raises` as a context manager).

Should solve https://github.com/earthlab/earthpy/issues/39

@lwasser have a look and see what you think!

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -13,6 +13,7 @@ from shapely.geometry import mapping, box
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from skimage import exposure
 
+
 def extent_to_json(ext_obj):
     """Convert bounds to a shapely geojson like spatial object.
     Helper function
@@ -30,13 +31,14 @@ def extent_to_json(ext_obj):
     """
 
     if type(ext_obj) == gpd.geodataframe.GeoDataFrame:
-        extent_json = mapping(box(*ext_obj.bounds.values[0]))
+        extent_json = mapping(box(*ext_obj.total_bounds))
     elif type(ext_obj) == list:
-        extent_json = mapping(box(ext_obj))
+        extent_json = mapping(box(*ext_obj))
     else:
-        raise ValueError("Please provide a geodataframe of a list of values - minx, miny, maxx, maxy")
+        raise ValueError("Please provide a geodataframe of a list of values.")
 
     return extent_json
+
 
 # calculate normalized difference between two arrays
 # both arrays must be of the same size
--- a/earthpy/tests/test_spatial.py
+++ b/earthpy/tests/test_spatial.py
@@ -1,6 +1,32 @@
-from earthpy import spatial
+import earthpy.spatial as es
+from shapely.geometry import Polygon, Point
+import geopandas as gpd
+import pandas as pd
+import pytest
 
 
-def test_dummy():
-    """"Dummy test that will pass."""
-    assert 1 == 1
+def test_extent_to_json():
+    """"Unit tests for extent_to_json()."""
+    # giving a list [minx, miny, maxx, maxy] makes a polygon
+    list_out = es.extent_to_json([0, 0, 1, 1])
+    assert list_out['type'] == 'Polygon'
+
+    # the polygon is the unit square
+    list_poly = Polygon(list_out['coordinates'][0])
+    assert list_poly.area == 1
+    assert list_poly.length == 4
+
+    # providing a GeoDataFrame creates identical output
+    df = pd.DataFrame(
+        {'lat': [0, 1],
+         'lon': [0, 1]}
+    )
+    df['coords'] = list(zip(df.lon, df.lat))
+    df['coords'] = df['coords'].apply(Point)
+    gdf = gpd.GeoDataFrame(df, geometry='coords')
+    gdf_out = es.extent_to_json(gdf)
+    assert gdf_out == list_out
+
+    # giving non-list or GeoDataFrame input raises a ValueError
+    with pytest.raises(ValueError):
+        es.extent_to_json({'a': 'dict'})


earthpy/spatial.py
```diff
@@ -30,14 +31,15 @@ def extent_to_json(ext_obj):
     """
 
     if type(ext_obj) == gpd.geodataframe.GeoDataFrame:
-        extent_json = mapping(box(*ext_obj.bounds.values[0]))
+        extent_json = mapping(box(*ext_obj.total_bounds))
```
mbjoseph: I think this `total_bounds` attribute is more general than `bounds.values[0]`, because it will work with points (GeoDataFrames with multiple rows) and polygons alike, but maybe you should make sure this still works for your intended use cases @lwasser

earthpy/spatial.py
```diff
@@ -30,14 +31,15 @@ def extent_to_json(ext_obj):
...
     elif type(ext_obj) == list:
-        extent_json = mapping(box(ext_obj))
+        extent_json = mapping(box(*ext_obj))
     else:
-        raise ValueError("Please provide a geodataframe of a list of values - minx, miny, maxx, maxy")
+        raise ValueError("Please provide a geodataframe of a list of values.")
```
mbjoseph: Shortened this error message so that we only have to maintain the order of list elements in one place (the docstring)

mbjoseph: Also this avoids us having to deal with handling multiline exceptions

earthpy/spatial.py
lwasser: oh you're right @mbjoseph that value makes no sense for a single point. (bounds) then again this is a note for me as what happens if this is fed a shapefile with only one point?!!! i'm not sure i thought of that edge case at all.

earthpy/spatial.py
lwasser: ok @mbjoseph  there is a typo here (my typo) it should say 

`"Please provide a geodataframe or a list of values."`

Ok let's think about this. this is just testing that the object is of one of the correct types. And i suppose if the min and max values are mixed up that is pretty hard to test for... so this makes sense! Just fix that one small typo pretty please but very good update here. i appreciate it!

earthpy/spatial.py
lwasser: agreed!!

missing edge case: if a user provides a single point geodataframe... what happens? It should fail gracefully. Should we provide it that in a test of some sort? i totally missed that edge case.

earthpy/tests/test_spatial.py
```diff
@@ -1,6 +1,32 @@
...
+import pandas as pd
+import pytest
 
 
-def test_dummy():
-    """"Dummy test that will pass."""
-    assert 1 == 1
+def test_extent_to_json():
+    """"Unit tests for extent_to_json()."""
+    # giving a list [minx, miny, maxx, maxy] makes a polygon
```
lwasser: We've been following rasterio and PEP 8 suggests capitalizing the first word of each comment line. Do you mind making that update unless you have a strong argument against it? Not all packages do it but i've been following Sean's lead as he seems to really know what he is doing and i appreciate how robust that package is.

earthpy/spatial.py
mbjoseph: Good question! If a user gives a single point GeoDataFrame, then they get a bounding "point" - basically a bounding box where the lower left and upper right corners are on the point. I would lean towards not raising an exception for this case, to keep this function as general as the geopandas backend (another way of saying that if geopandas allows it, we might as well allow it too).

mbjoseph: One other bit of weirdness that I will fix though is that `shapely.geometry.box` does not raise exceptions when minx is greater than maxx, or when miny is greater than maxy. For that, I think it would make sense to add an exception to help users out in case they mix up x and y.

earthpy/spatial.py
mbjoseph: Typo is fixed with 4bcaca912f2f61a04547e83525d81db63f2e45e4 - I would recommend in the future that we try to use duck typing rather than testing object types unless absolutely necessary, to get extra cool "Pythonic" street cred.

earthpy/spatial.py
lwasser: Right but we wrote this function to support the rasterio.mask() function whic requires a geojson object. So when you mask you will get a really weird output?? maybe a point. i dont even know. Could we raise a warning perhaps? OR should that be raised in crop_image() instead? maybe i'm thinking ahead of myself in terms of workflows and not seeing this as a basic general function! it's happened before.:)

lwasser: sure @mbjoseph do you mean in case they mix up min and max (as opposed to x and y)? it would be tricky i think to test for x vs y  :) but perhaps i dont understand! Does shapely just fix the order? i mean we *could* maybe do that too? it's all risky business however with boundaries as they are just numbers. the bounds object is a bit more explicit.

mbjoseph: Hmmmmm - that's worth thinking about. I think if we can decide how we want to handle this case, then we can figure out whether we want to raise an error, warning, or not. What's the behavior that you would want if you were masking a raster with a point? 

For me I would maybe expect to get one pixel back, with everything else being masked out (or if `invert=True`, everything except for that pixel). I have not tested this with `rasterio.mask` to see what happens in this case.

lwasser: i haven't either. i just hadn't thought about this at all until you brought this up here! i think we can have the conversation when we get to` crop_image` however as you are right. this just takes a geometry and makes a dictionary / geojson like thing. so that is what it does and it doesn't have to ask questions relative to other applications (like `mask()` . i'm complicating life here with my questions.

earthpy/spatial.py
lwasser: sure but for this particular instance, can you show me how duck typing would be applied? this is not a class, it's logic to determine how to process the output. I do love me some python street cred tho!

earthpy/spatial.py
mbjoseph: Sounds good! I'm glad we're exploring these edge cases, and when we get to testing `crop_image`, we should definitely keep this one in mind.

earthpy/tests/test_spatial.py
mbjoseph: Bah oops. Fixed with 09a5dbaf9f50179859b5bab53cae4ce6b396a156

mbjoseph:
ok @lwasser I think this PR is good to merge at this point! I added a note to our future selves to handle the edge cases when testing the crop function: https://github.com/earthlab/earthpy/issues/34

earthpy/spatial.py
mbjoseph: Word! I think in this case it's fine to have it as written, but there is a slight loss of generality. For instance, if a user wanted to pass in a tuple of xmin, ymin, xmax, ymax, there's not much reason to raise an error assuming that those can be unpacked in the `box` function, and that we can check that `xmin <= xmax` and `ymin <= ymax`.

lwasser:
omg don't shoot me @mbjoseph  i'm a PEP 8 stickler. PEP 8 suggests ordering packages from most common to least common and thus earthpy would go last. 

Often also what is suggested is then you alphabetize after that. [See Contributing PR](https://github.com/earthlab/earthpy/pull/47). I"ll add more about those things that I think people will miss and that PSSSS i haven't fixed this in the main code body because the students should be able to catch this stuff as i've been on them all semester about it! But i'd love our test scripts to follow suite. I promise i'm not trying to be difficult tonight. 

Please change this:
```
import earthpy.spatial as es
from shapely.geometry import Polygon, Point
import geopandas as gpd
import pandas as pd
import pytest
```
to:

```
import pandas as pd
import pytest
from shapely.geometry import Polygon, Point
import geopandas as gpd
import earthpy.spatial as es
```
I promise last nitpicky thing. .. i hope. i'll add the comment syntax and such to the contributing file now. I get it PEP 8 is kinda ginormous to wrap your head around. i just happen to have spare time now w this arm of mine.

mbjoseph:
No worries! For some reason I think my linter is not working... But this should be fixed with c13bdbd

lwasser:
I’d love to know how you lint. I have been lazily using jupyter notebooks for everything these days but was trying to figure out how to lint in pycharm. What do you use?

Also the capitalization and a few other things aren’t handled by autopep8 so I’m not sure they are always handled in general. But if you did into the docs for PEP 8 you’ll see that stuff.

lwasser:
Sweet. That’s what I use! But don’t have the linters setup. I’ll look for PEP 8 !! I often code in Atom these days too. Kinda weird but works for me.

earthpy/tests/test_spatial.py
```diff
@@ -1,6 +1,39 @@
...
+    # Giving a list [minx, miny, maxx, maxy] makes a polygon
+    list_out = es.extent_to_json([0, 0, 1, 1])
+    assert list_out['type'] == 'Polygon'
+
+    # The polygon is the unit square
+    list_poly = Polygon(list_out['coordinates'][0])
+    assert list_poly.area == 1
+    assert list_poly.length == 4
+
+    # Providing a GeoDataFrame creates identical output
```
lwasser: Should we make a geodataframe builder function @mbjoseph that accepts a list of points and created a gdf? i can create an issue. i had this thought last night.

mbjoseph: Sure! This seems like a useful utility to have in our test suite. I can think of at least a few places where it would be useful:

- here, to make a GeoDataFrame that has some extent
- to test the `crop` function edge cases
- if we ever do anything with rasterizing points or drawing polygons around points

PR #84, opened 2018-10-30
Update spatial.py to conform to PEP-8 standards, update docstring returns

### LINKED ISSUE #74: Update spatial.py to conform to PEP-8 standards and update returns in spatial.py docstrings
## Identify the Documentation Element That You Propose to Update
1.  Update spatial.py to conform to PEP-8 standards.
2.  Update several docstrings in spatial.py to include/update returned variables.

### Proposed change 1
a. Remove spaces around keyword assignments and adding spaces after commas on the following lines: 138, 170, 174, 175, 250, 282, 332, 338, 339, 340, 341, 342, 343, 344, 393, 394, 401, 402, 403, 404.
b.  Fix indentation on the following lines: 131, 132, 133, 134, 135, 301, 321, 322, 393, 425, 444. 
c.  Break up lines that are greater than 120 characters (in a couple of locations).
d.  Remove an unused import `from rasterio.mask import mask`

### Proposed change 2
a.  Function docstrings that need a description of the return added: `normalized_diff()`, `stack_raster_tifs()`.  
b.  Function docstrings that need their return description updated: `plot_bands()` (Function doesn't actually return anything.  Should this return the plt and/or ax object?), `plot_rgb()` (Function doesn't actually return anything.  Should this return the ax object as described?), `hist()` (Function doesn't actually return anything.  Should this return the plt and/or ax object?)

mikedorfman:
Resolves #74 

@lwasser Please let me know if you have any questions.

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -2,7 +2,6 @@ import contextlib
 import os
 import geopandas as gpd
 import rasterio as rio
-from rasterio.mask import mask
 import numpy as np
 import numpy.ma as ma
 import matplotlib.pyplot as plt
@@ -52,6 +51,11 @@ def normalized_diff(b1, b2):
     ----------
     b1, b2 : arrays with the same shape
         Math will be calculated (b2-b1) / (b2+b1).
+
+    Returns
+    ----------
+    n_diff : ndarray with the same shape as inputs
+        The element-wise result of (b2-b1) / (b2+b1) with all nan values masked.
     """
     if not (b1.shape == b2.shape):
         raise ValueError("Both arrays should be of the same dimensions")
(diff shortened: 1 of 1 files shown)


earthpy/spatial.py
```diff
@@ -73,12 +77,22 @@ def stack_raster_tifs(band_paths, out_path, arr_out=True):
...
 
+    Returns
+    ----------
+    If arr_out keyword is True:
+        tuple: The first value representing the result of src.read() of the stacked array and the second value
+        representing the result of src.profile of the stacked array.
+    If arr_out keyword is False:
+        str : A path with a file name for the output stacked raster tif file.
+
+    TODO: Instead of returning a file path when arr_out=False, consider returning None since the out_path is already
```
lwasser: Good recommendation @mikedorfman I will look into that. You are correct in that the user does know that path. Would you want to see a message like "file created" ? We could have `verbose =` as a boolean? Or would you rather it write the file silently?

earthpy/spatial.py
```diff
@@ -390,18 +408,21 @@ def plot_rgb(arr, rgb = (0,1,2),
 
     # Then plot. Define ax if it's default to none
     if ax is None:
-      fig, ax = plt.subplots(figsize = figsize)
-    ax.imshow(rgb_bands, extent = extent)
+        fig, ax = plt.subplots(figsize=figsize)
+    else:
+        fig = None
```
lwasser: cool . i haven't set `fig` to `None` before... so it will be interesting to test this out. This is much cleaner as written now as it's more explicit.

earthpy/spatial.py
mikedorfman: @lwasser I feel like writing the file silently makes the most sense, although I wouldn't be opposed to a verbose keyword.  If you do this, you can simplify lines 114-124 to be:
`
# Write stacked gtif file
with rio.open(out_path, 'w', **dest_kwargs) as dest:
    stack(sources, dest)

if arr_out == True:
    # Read and return array
    with rio.open(out_path, 'r') as src:
        return(src.read(), src.profile)
return None
`

mikedorfman:
Ok, sounds good!  Thanks @lwasser!

lwasser:
Testing looks good! This all looks good @mikedorfman and all tests are passing so i'm going to merge this. Thank you again and congrats on your first contribution to earthpy. We appreciate it!

Issue #2, opened 2018-04-18
Normalized diff function - What order would a user expect the bands to be in?


lwasser:
Currently we have  a `normalized_diff()` function which calculated norm diff of 2 bands. 
It's nice and generic in that it can be used on ANY raster data BUT question

1. is the band order what a user would expect? i think it's the OPPOSITE of what i taught in my R class. 
2. the name could be shorter. SHould i rename it to `norm_diff()` ? 

hey @mbjoseph @joemcglinchy @jlpalomino have a look at the function and see what you think. I won't change it during the semester but am open to changing it in December if you guys think it should be the opposite? Or is it ok?

joemcglinchy:
https://github.com/earthlab/earthpy/blob/9a4fc7cd765152b727a8d59834623d72f0ac9d1e/earthpy/spatial.py#L43

I think the doc is perfect, but the order of the arguments should be reversed. I think this is more intuitive since b2 comes first in the numerator and denominator for the normalized difference equation.

mbjoseph:
I wouldn't have any strong prior expectations for the ordering of bands for a function like this.

joemcglinchy:
it is more subjective... as long as the doc is clear on which bands go where in the equation, which I think it is. seeing the equation (b2 - b1) / (b2 + b1), implies to me, that the the args should come in as b2 and then b1, but that's a little weird since it sounds like you're counting backwards and we read left to right.

lwasser:
just a note all that in line with @joemcglinchy 's original thought and @jlpalomino 's second.... that the order of bands in the norm_diff function is confusing... several students have voiced the same issue! This is why i originally created this issue. @joemcglinchy i know you made some changes... Do you want to resubmit the PR to adjust the function and also update the docs, [here](https://github.com/earthlab/earthpy/blob/master/docs/spatial-raster.rst) and we can then decide when to merge it the students have an assignment where they will likely use it starting today. but after thanksgiving could be a good time to merge this. then we'd want to update our lessons too. not a big deal to do tho!

lwasser:
@jlpalomino  assigning this issue to you when you add tests for norm_diff please also just swap the parameter order!

lwasser:
referencing this pr which has the changes implemented.  - https://github.com/earthlab/earthpy/pull/140

@jlpalomino will you please also close this issue when the PR is merged!

PR #90, opened 2018-10-30
Spatial.py update imported libraries to pep8, Edited documentation to pep8, suggested function edit

### LINKED ISSUE #54: Spatial.py edit documentation and import libraries to pep8 guidelines; suggest code edit for normalized_diff & stack_raster_tifs
## Identify the Documentation Element That You Propose to Update
Edit the documentation to match pep8 guidelines and rearrange the import libraries to follow pep8 guidelines wherein more general libraries are listed first and more specific libraries after. 

Also delete comments that seem unnecessary, and add comments to suggest a change to normalized_diff function to make it less confusing 

## Please Identify Your Proposed Changes Below

**Documentation should begin with a space and follow with a capital letter (i.e ' # Comment here')**

**Importing should begin with the most general libraries first**

**Delete comments such as**
1.)  #ndvi[np.isnan(ndvi)] = 0 *: commented out in normalized_diff function*
2.) # Should this wrap around show instead of plotting as it does? *: commented in plot_rgb*
3.)  #ax.set(xticks=[], yticks=[]) *:commented out in the hist function*

**Add comments**
1.) For normalized_diff function to suggest a change ( # Could be confusing if one does not know that b2 is subtracted from b1. 
#An individual might assume the first input is subtracted from the second. 
#For example, it could be more commonly assumed : (b1-b2)/(b1+b2) ) 

2.) Add comment for stack_raster_tifs function to include .sort()

shwh2628:
Editted documentation to pep8 guidelines (# Comment here) and deleted unnecessary code that was commented out. 

Included suggestions for code (normalized_diff and stack_raster_tifs) 

Arranged imported libraries to pep8 guidelines. 

Fix #54

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -1,12 +1,12 @@
-import contextlib
 import os
 import geopandas as gpd
-import rasterio as rio
 import numpy as np
 import numpy.ma as ma
 import matplotlib.pyplot as plt
 from matplotlib import patches as mpatches
-
+import rasterio as rio
+from rasterio.mask import mask
+import contextlib
 from shapely.geometry import mapping, box
 # For color bar resizing
 from mpl_toolkits.axes_grid1 import make_axes_locatable
@@ -43,6 +43,9 @@ def extent_to_json(ext_obj):
 
 # Calculate normalized difference between two arrays
 # Both arrays must be of the same size
+# This could be confusing if one does not know that b2 is subtracted from b1. 
+# An individual might assume the first input is subtracted from the second. 
+# For example, it could be more commonly assumed : (b1-b2)/(b1+b2)
 def normalized_diff(b1, b2):
     """Take two numpy arrays and calculate the normalized difference
     Math will be calculated (b2-b1) / (b2+b1).
@@ -61,12 +64,12 @@ def normalized_diff(b1, b2):
         raise ValueError("Both arrays should be of the same dimensions")
 
     n_diff = (b2 - b1) / (b2 + b1)
-    #ndvi[np.isnan(ndvi)] = 0
     n_diff = np.ma.masked_invalid(n_diff)
     return n_diff
 
 
-# TODO: include a no data value here if provided ...
+# TODO: include a no data value here if provided 
+# TODO: Include .sort() in the function
 def stack_raster_tifs(band_paths, out_path, arr_out=True):
     """Take a list of raster paths and turn into an ouput raster stack in numpy format.
     Note that this function depends upon the stack() function.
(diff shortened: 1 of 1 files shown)


lwasser:
hey @shwh2628  thank you for this PR. in the future please be sure to tag whomever you are working with when you submit a PR to get their attention. in my case @lwasser would work! Reviewing this now.

earthpy/spatial.py
```diff
@@ -1,12 +1,12 @@
-import contextlib
 import os
 import geopandas as gpd
```
lwasser: following proper order numpy, and os and contextlib should go at the top. Then matplotlib.  Then the spatil stuff then earthpy. please adjust the order @shwh2628  to follow conventions! geopandas is definitely considered a 3rd party package!

earthpy/spatial.py
```diff
@@ -43,6 +43,9 @@ def extent_to_json(ext_obj):
 
 # Calculate normalized difference between two arrays
 # Both arrays must be of the same size
+# This could be confusing if one does not know that b2 is subtracted from b1.
```
lwasser: @shwh2628 it is not appropriate to place additional commentary into a file. the goal is to REMOVE comments from a file like this not add more. 

If you wanted to adjust the structure of the function, you should submit a PR suggesting the band order be swapped in the code. Please REMOVE this change. 

this wold be a change you would have a discussion about with the package maintainers and if you looked closely at the issue you would see that discussion 

[here](https://github.com/earthlab/earthpy/issues/2) as an issue!

earthpy/spatial.py
```diff
@@ -61,12 +64,12 @@ def normalized_diff(b1, b2):
...
 
     n_diff = (b2 - b1) / (b2 + b1)
-    #ndvi[np.isnan(ndvi)] = 0
     n_diff = np.ma.masked_invalid(n_diff)
     return n_diff
 
 
-# TODO: include a no data value here if provided ...
+# TODO: include a no data value here if provided 
+# TODO: Include .sort() in the function
```
lwasser: this is another example of making a suggestion that the developers should approve. In this case i would argue that sort should happen after glob and before being passed to this function. This function's intent is not to order lists but rather to create a stack. It assumes the user passes what it wants. Please remove this  line frm your PR. 

A todo should be a conversation with the developers that they
1. agree to do and then
2. ideally your pr would fix that issue. 
Here i disagree that this is a todo.

shwh2628:
Hi @lwasser , I (think) I pushed another commit. 

Did it work??

lwasser:
it sure did. this looks good @shwh2628 !! i will merge it later today once the checks pass (i know they will) .  i'll ping you when i merge it :)

lwasser:
hi @shwh2628 all tests have passed! i will merge this now! Thank you so much for this PR and congratulations on your first contribution to earthpy!

Issue #75, opened 2018-10-28
Clean clip.py code AND capitalize comment on line 322 of spatial.py code


tkarfs1:
## Identify the Documentation Element That You Propose to Update
1. Clean clip.py code using Pep8 style guide
2. Line 322 in spatial.py is not capitalized

## Please Identify Your Proposed Changes Below
1.  Delete whitespace and improve indent
2. Capitalize markdown line 299 "test" to Test"

lwasser:
@tkarfs1 i'm assigning this to @mbjoseph but did want to give you a heads up that i believe spatial.py has been worked on already for PEP 8 and there is a second proposal to work on it here.

https://github.com/earthlab/earthpy/issues/74

so you will need to submit a separate fix for item 1. 

To see that this work was done - please update your fork. We will need you to submit another fix given my comment above. I believe no one has proposed to document plot_bands so you should be good on your second proposed item. 
I am assigning this issue to my good colleague @mbjoseph !

tkarfs1:
Hi @lwasser @mbjoseph can I still capitalize "test" in line [322](https://github.com/earthlab/earthpy/blob/c325d4a07d80df0275c278d1400fde285faf99a7/earthpy/spatial.py#L322)

mbjoseph:
Hi @tkarfs1! Yes that is still a valid and needed change! 

I believe @lwasser wants you to find a second change/fix to propose before moving forward.

mbjoseph:
Hello @tkarfs1 - yes that is a valid change that still needs doing. However your second change, to document `plot_bands()`, has already been fixed (see the most recent documentation here: https://github.com/earthlab/earthpy/blob/master/docs/spatial-raster.rst#plot-raster-file-bands). 

Can you propose an alternative second change to make? 

Thanks!

tkarfs1:
Thanks @mbjoseph ! I'll also apply pep 8 standards to the clip.py code. It looks like I can delete some white space in that notebook.

mbjoseph:
Ok sounds good @tkarfs1. When you submit your pull request please "@" me by including @mbjoseph!

tkarfs1:
This change has been implemented via pull request BUT there seems to be a conflict with the base in relation to deleted whitespace in clip.py

mbjoseph:
Nice job @tkarfs1! Thanks for the pull request

PR #93, opened 2018-10-30
Deleted whitespace in clip.py AND Capitalized "Test" in spatial.py

tkarfs1:
I thought there was more whitespace to be removed but Pep8 standards said otherwise! @mbjoseph


mbjoseph:
Solves #75

PR #101, opened 2018-11-03
[WIP] - Developing tests for the clip module 

lwasser:
ok @mbjoseph  here is a very cursory start at some tests for the clip function. i'm not exactly sure what tests we need but in creating the tests i added assertions and such to the functions too!

Trying to use ducktyping (quack) when possible and assuming that some things - like CRS are actually tested by geopandas so for instance .intersects fails if the CRS' are different. When you have time to give this a review please do so. it can be next week to!!


lwasser:
questions
1. Do tests run automagically on travis once added to the tests dir?
2. Given we may have to create several different GeoDataFrames ... do we want all of that mess in the tests file or is there a cleaner way? I do plan to create some weirdo geometries like donuts w holes and such! just gonna get cluttered doing that.
3. i have no idea how to create rasters yet beyond creating a numpy array... so starting with vectors which are easy peasy.

Issue #26, opened 2018-10-18
draw_legend is not ready to use


lwasser:
Currently there is a draw_legend function that is supposed to make creating a nice legend for a plot easier. However as it is built, it's not general enough to use consistently.

I think i should remove it for the time being until it's fully tested but just noting here that right now it's in the package and really does not work.

lwasser:
Adding a noteo from @mbjoseph  

draw_legend code is fairly complex as written with a bunch of nested list comprehensions so as we fix it also consider simplifying the code.

lwasser:
closing this issue as it is fixed with this pr https://github.com/earthlab/earthpy/pull/149

PR #108, opened 2018-11-03
Tests for clip module 

lwasser:


earthpy/clip.py
```diff
@@ -2,6 +2,7 @@
 
 "A module to clip vector data using geopandas"
 
+# TODO: Clip poly should use OVERLAY not spatial indexing + intersects
```
mbjoseph: @lwasser should this be an issue on the GitHub repo instead of a comment?

earthpy/clip.py
```diff
@@ -83,34 +84,45 @@ def clip_line_poly(shp, clip_obj):
...
+        shp.geometry
+        clip_obj.geometry
+    except (AttributeError, TypeError):
+        raise AssertionError('Input variables should be GeoDataFrames with a geometry column')
+
+    # Test to ensure shapes intersect
+    if not any(shp.intersects(clip_obj)):
+        raise ValueError("Shape and crop extent do not overlap.")
+
     if shp["geometry"].iloc[0].type == "Point":
```
mbjoseph: @lwasser is it possible in geopandas to have a data frame where some rows (e.g., the first) contain points, and other rows contain lines or polygons?

earthpy/spatial.py
```diff
@@ -80,8 +81,8 @@ def normalized_diff(b1, b2):
 # TODO: include a no data value here if provided
 
 def stack_raster_tifs(band_paths, out_path, arr_out=True):
-    """Take a list of raster paths and turn into an output raster stack in numpy format.
-    Note that this function depends upon the stack() function.
+    """Take a list of raster paths and turn into an output raster stack in
+    numpy format. Note that this function depends upon the stack() function.
```
mbjoseph: Would it be more precise to say "as a numpy array" instead of "in numpy format"?

earthpy/spatial.py
```diff
@@ -209,7 +220,9 @@ def bytescale(data, cmin=None, cmax=None, high=255, low=0):
     Byte scaling means converting the input image to uint8 dtype and scaling
     the range to ``(low, high)`` (default 0-255).
     If the input image already has dtype uint8, no scaling is done.
-    This function is only available if Python Imaging Library (PIL) is installed.
+    This function is only available if Python Imaging Library (PIL) is
```
mbjoseph: Is this true? I can't see below how PIL is used.

earthpy/spatial.py
```diff
@@ -346,8 +362,8 @@ def plot_bands(arr, title=None, cmap="Greys_r", figsize=(12, 12), cols=3, extent
         # Test if there are enough titles to create plots
         if title:
             if not (len(title) == arr.shape[0]):
-                raise ValueError("The number of plot titles should be the same " +
-                                 "as the number of raster layers in your array.")
+                raise ValueError("The number of plot titles should equal " +
+                                 "as the number of array raster layers.")
```
mbjoseph: Delete "as"? Otherwise we have "The number of plot titles should equal as the number of array raster layers".

earthpy/spatial.py
```diff
@@ -432,7 +450,8 @@ def plot_rgb(arr, rgb=(0, 1, 2),
         arr_rescaled = np.zeros_like(rgb_bands)
         for ii, band in enumerate(rgb_bands):
             p2, p98 = np.percentile(band, (s_min, s_max))
```
mbjoseph: The names `p2` and `p98` only make sense for the default value of `str_clip`. Why not call these `low` and `high` or `lower_lim` and `upper_lim`?

earthpy/spatial.py
```diff
@@ -441,7 +460,9 @@ def plot_rgb(arr, rgb=(0, 1, 2),
         mask = ~(np.ma.getmask(rgb_bands[0])) * 255
 
         # Add the mask to the array (ise earthpy bytescale)
-        rgb_bands = np.vstack((bytescale(rgb_bands), np.expand_dims(mask, axis=0))).transpose([1, 2, 0])
+        rgb_bands = np.vstack((bytescale(rgb_bands),
+                               np.expand_dims(mask, axis=0))).\
+            transpose([1, 2, 0])
```
mbjoseph: A comment here to explain this magic number sequence `[1, 2, 0]` would be very useful!

earthpy/spatial.py
```diff
@@ -457,20 +478,23 @@ def plot_rgb(arr, rgb=(0, 1, 2),
...
     """
-    Plot histogram each layer in a raster stack converted into a numpy array for quick visualization.
+    Plot histogram for each layer in a numpy array.
 
     Parameters
     ----------
     arr: a n dimension numpy array
-    title: a list of title values that should either equal the number of bands or be empty, default = none
-    colors: a list of color values that should either equal the number of bands or be a single color, (purple = default)
+    title: str_clip
```
mbjoseph: Is this supposed to be `str_clip`?

earthpy/spatial.py
```diff
@@ -573,7 +601,8 @@ def draw_legend(im, classes, titles, bbox=(1.05, 1), loc=2):
     colors = [im.cmap(im.norm(aclass)) for aclass in classes]
 
     patches = [mpatches.Patch(color=colors[i],
-                              label="{l}".format(l=titles[i])) for i in range(len(titles))]
+                              label="{l}".
+                              format(l=titles[i])) for i in range(len(titles))]
```
mbjoseph: This 3 line statement has a lot going on. Possible to break it up somehow and improve clarity with good names?

earthpy/tests/test_spatial.py
```diff
@@ -37,3 +40,100 @@ def test_extent_to_json():
...
+donut_geom = gpd.overlay(locs_buff, poly_in_gdf,
+                         how='symmetric_difference')
+
+multi_poly = donut_geom.unary_union
+multi_gdf = gpd.GeoDataFrame(gpd.GeoSeries(multi_poly),
+                   crs={'init': 'epsg:4326'})
+multi_gdf = multi_gdf.rename(columns={0: 'geometry'}).\
+    set_geometry('geometry')
+
+# LINE OBJECTS
```
mbjoseph: ?

earthpy/tests/test_spatial.py
```diff
@@ -37,3 +40,100 @@ def test_extent_to_json():
...
+
+def test_clip_funs():
+    # Test that a bounding box returns error if the extents don't overlap
```
mbjoseph: You can break up this function into multiple smaller functions that have good names, and then you don't need so much narrative in the comments, e.g., 

`test_nonoverlapping_extents_error()`, `test_geometries_error()`, `test_clip_point()`, `test_clip_poly()`, `test_clip_multipoly()`, `test_clip_donut()`, `test_clip_line()`

If you want to have comments for these smaller functions, the docstrings are probably a good place! Breaking this out into multiple functions will also allow you to see which ones pass and fail, rather than having the whole thing fail if any one assertion fails.

earthpy/clip.py
lwasser: @mbjoseph i'm fairly certain the answer for a GeoDataFrame is no given it's based on the shapefile structure. The idea is that a single object contains one data type - points OR lines or Polys. Thus my guess is if you try to feed a GDF both it will be super confused. NOW with that said i suppose it is possible to have a line that is actually a POLY but created as a line object. Think county boundaries for instance. sometimes that will be a line feature even though it should be a poly.  Are you suggesting another test for something like that? Or just ensuring == "Point" is always true! ?

earthpy/spatial.py
lwasser: yes ! much more precise !

earthpy/spatial.py
lwasser: oh totally not true!

earthpy/spatial.py
lwasser: do you think using this https://github.com/mapbox/rasterio/blob/147df8a2356d6c59188eeaa621e7e67b8a764e31/rasterio/plot.py#L184 `reshape_as_image()` function from rasterio would make the code better / more expressive vs a comment? i've been thinking about doing that anyway. although it's pretty clean to just do a .transpose but not expressive at all.

earthpy/spatial.py
lwasser: goodness that must have been atom autocomplete. Str.  fixed.

earthpy/spatial.py
lwasser: i don't think we can easily break it up .. maybe i could pull out some of the nested list comprehensions tho!? 

welcome to the backend of matplotlib. but we could def create better var names! This is creating a legend. each patch is a square in the legend. each title is a legend "category" think about say a landuse map with categories like agriculture, urban, etc. each of those words is a title. each box has a color associated with that category, etc. hmmm lemme think about this. this function needs to be rewritten as it actually doesn't work now.

earthpy/tests/test_spatial.py
lwasser: These comments were me trying to sort out what sets of functions we might consider for gdf creation. I created points, lines and polys and wants to keep the creation of each separate in the code in hopes it would be easy to identify how it could be better modularized! was hoping you'd see that and maybe address whether we should consider functions to create gdfs.

earthpy/tests/test_spatial.py
lwasser: ok great. this was another question i had for you. answered here. i'd prefer to break it out like that just wasn't sure if the goal was a single function!

earthpy/clip.py
mbjoseph: Ah cool @lwasser! I wasn't sure whether just checking the first element of the 'geometry' series was going to be robust enough. But, it sounds like if the first element is a "Point", then the rest must be.

earthpy/spatial.py
mbjoseph: Either way seems fine to me, and less mysterious!

earthpy/spatial.py
mbjoseph: Ugh matplotlib... :nauseated_face:

earthpy/tests/test_spatial.py
mbjoseph: This one in particular seemed weird, since I couldn't find any line objects below it!

mbjoseph: We could totally write functions to generate these geodataframes, especially if we need to do it elsewhere.

lwasser:
Ok goodness. I was working on this a bit later last night so prob just got tired and messed up the organization! The lines are there somewhere 😊 Anyway sounds good. I’ll work on a function and can use it in your tests for extent_to_json as well!!!

earthpy/clip.py
lwasser: it should but also it's how i personally track code i'm working on! it's nice and bright green in pycharm !! yes TODO is a thing! So i actually want to keep it here as well. issues in github are super useful but i appreciate these types of comments in the code too.

earthpy/clip.py
lwasser: Yes -- i would assume that would be the case all of the time given the rules of "shapefile" and GDF structure. If it was a geodatabase i would not assume that. it's a great thing to check tho - i appreciate the comment!

earthpy/spatial.py
lwasser: fixed!!

earthpy/spatial.py
lwasser: fixed!

earthpy/spatial.py
lwasser: fixed and also made this a """double line string here"""

earthpy/spatial.py
lwasser: fixed!!

earthpy/spatial.py
lwasser: added a comment for the time being!!

earthpy/spatial.py
lwasser: Because this is a can of worms - i've added this note to the [issue here](https://github.com/earthlab/earthpy/issues/26) to ensure it is addressed. I don't think i can address it in this PR however. matplotlib is such a beast. i.cant.even.

earthpy/tests/test_spatial.py
lwasser: ok cool. i'll work on cleaning up the code first with smaller expressive functions. Then maybe we create functions for GDF creation.

mbjoseph:
Good morning @lwasser!

>    1. How do i test the structure of the returned object with ducktyping or just use isinstance()?

The test as written above is going to pass only if `cl.clip_shp` raises an `AssertionError`. I'm not totally sure what this test is supposed to test for. The name of the function makes me think you want to ensure that a GDF gets returned. The :duck: typing way AFAIK would be to assert that whatever is returns quacks like a GDF, e.g., 

```python
def returns_gdf():
    out = cl.clip_shp(locs, poly_in_gdf)
    assert hasattr(out, 'geometry')    
```

But the docstring makes me think you want to test that the function returns an assertion error when given two inputs that are not GDFs, which would be a different test, e.g., 

```python
def test_error_non_gdf_args():
    """"Test that clipping fails for non GeoDataFrame arguments.

    Clipping should fail regardless of order and number of invalid args.
    """
    with pytest.raises(AssertionError):
        cl.clip_shp(not_a_gdf, is_a_gdf)
    with pytest.raises(AssertionError):
        cl.clip_shp(is_a_gdf, not_a_gdf)
    with pytest.raises(AssertionError):
        cl.clip_shp(not_a_gdf, not_a_gdf)
```

>    2. Should the clip tests be in a separate tests_clip.py file?

Yes - I think this is a good idea. Having the structure of the test files mirror the structure of the source files will make it easier to find and add tests.

PR #115, opened 2018-11-07
Updated tests for the clip_shp function

lwasser:


lwasser:
dangit -- [be sure to address this](https://github.com/earthlab/earthpy/pull/108#issuecomment-436662864) since i did some magic at the command line and by MISTAKE merged the other PR.

lwasser:
im sorry @mbjoseph i had good intentions to break out sets of changes into two pr's and instead managed to merge BOTH at the CLI. :) still working on this  however :) luckily it's just us that see it so i'm not going to try to undo what i did...

lwasser:
Note: through building unit tests i've found a flaw in clipping behavior. `.intersection()`, when provided a line that has a SINGLE POINT that exactly overlaps with the extent of the clipping object does not clip that line string. When it has a single point that overlaps and another point where it could create a line segment, you get a multi-line. so need to dig a bit into this more as this could happen!

earthpy/tests/test_clip.py
```diff
@@ -24,7 +23,7 @@
 """ Create Line Objects For Testing """
 
 linea = LineString([(1, 1), (2, 2), (3, 2), (5, 3)])
-lineb = LineString([(3, 4), (5, 7), (12, 2), (10, 5), (17, 17)])
+lineb = LineString([(3, 4), (5, 7), (12, 2), (10, 5), (9, 7.5)])
```
lwasser: here is the fix. if the  last set of points in lineb is outside of the clip extent then the entire line is NOT clipped.

lwasser:
oh heyyyyy @mbjoseph guess what i just fixed this - i think! i'm not sure why my coverage dipped a bit below 100% here but... this should cover most of our bases.

NOTE: there is[ one outstanding issue see here](https://github.com/geopandas/geopandas/issues/861)

But i can't address this until our geopandas colleagues provide some insight into... 

wait for it...
creating a line that happens to have ONE VERTEX that exactly overlaps the clipping geometry! Oh yes. 

i by chance made that happen in a test and found this issue. So i *think* for now we can merge this for real but wouldn't mind one more review from you! lemme know whassss goooodddd !!! (or not) 

#codelyfe

lwasser:
yeee hawwww!!! thanks @mbjoseph merging this.

PR #116, opened 2018-11-08
updated mask.py to use np.isin() for mask creation

joemcglinchy:
I verified the mask creation with np.isin() works to create the array of 1's and 0's, 1's being masked entries. I also fixed the `make_apply_mask` method to call the appropriate arguments. I think it should be good to go!

### DIFF
--- a/earthpy/mask.py
+++ b/earthpy/mask.py
@@ -22,10 +22,10 @@ def make_cloud_mask(mask_arr, vals):
         True (Boolean)
     """
 
-    # construct the mask
+    # Construct the mask
     temp_mask = np.isin(mask_arr, vals)
     
-    # mask the values
+    # Mask the values
     mask_arr[temp_mask] = 1
     mask_arr[~temp_mask] = 0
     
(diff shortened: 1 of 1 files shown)


earthpy/mask.py
```diff
@@ -21,11 +22,13 @@ def make_cloud_mask(mask_arr, vals):
         True (Boolean)
     """
 
-    # Improved by Joe! Not tested yet but should work!
+    # construct the mask
```
lwasser: So much cleaner now! @joemcglinchy will you capitalize each comment line for PEP 8 standards so we're consistent?

lwasser:
@joemcglinchy what do you think about a dict lookup for landsat 7 and 8 and maybe modis? or just leave as is for now? this could be another issue where we discuss adding that AFTER we build some tests around these functions.

joemcglinchy:
> @joemcglinchy what do you think about a dict lookup for landsat 7 and 8 and maybe modis? or just leave as is for now? this could be another issue where we discuss adding that AFTER we build some tests around these functions.

I think we can leave as-is for now... I was wondering where to put that dictionary. Although! I could place the dictionary I am using currently into that file, and then to use it you could say `from mask import pixel_flags`, which is what I am doing on my own right now. So your import statement to use the functionality would look like `from mask import pixel_flags, make_apply_mask`. what do you think?

lwasser:
thank you @joemcglinchy

PR #122, opened 2018-11-08
[tests] plot_band tests - 100% coverage yay!!

lwasser:
let's see if this works!

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -346,24 +346,22 @@ def plot_bands(arr, title=None, cmap="Greys_r",
     --------
     >>>import earthpy.spatial as es
     ...
-    ...titles = ["Red Band", "Green Band", "Blue Band",
-    ...          "Near Infrared (NIR) Band"]
+    ...im = np.random.randint(10, size=(2, 4, 5))
+    ...titles = ["Red Band", "Green Band"]
     ...
     ...# Plot all bands of a raster tif
-    ...es.plot_bands(naip_image,
+    ...es.plot_bands(im,
     ...              title=titles,
     ...              figsize=(12,5),
     ...              cols=2)
     """
+    if not isinstance(arr, np.ndarray):
+        raise AssertionError("Input arr should be a numpy array")
 
     if (arr.ndim == 2) and (len(title) > 1):
         raise AssertionError("""You have provided more than one title for a
                              single band array""")
 
-    try:
-        arr.ndim
-    except AssertionError:
-        raise AssertionError("Input arr should be a numpy array")
 
     # If the array is 3 dimensional setup grid plotting
     if arr.ndim > 2 and arr.shape[0] > 1:
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -319,19 +319,20 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
 # Function to plot all layers in a stack
 def plot_bands(arr, title=None, cmap="Greys_r",
                figsize=(12, 12), cols=3, extent=None):
-    """Plot each layer in a raster stack converted into a numpy array for
-       quick visualization.
+    """Plot each layer in a raster stack read from rasterio in
+    (band, row ,col) order as a numpy array. plot_bands will create an
+    individual plot for each band in a grid.
 
     Parameters
     ----------
     arr: numpy array
-        An n-dimensional numpy array
+        An n-dimensional numpy array in (band, row, col) order
     title: str or list
         Title of one band, or list of titles with one title per band
     cmap: str
         Colormap name ("greys" by default)
     cols: int
-        Number of columns for plot grid (3 by default)
+        Number of columns for plot grid (default: 3)
     figsize: tuple - optional
         Figure size in inches ((12, 12) by default)
     extent: tuple - optional
@@ -389,26 +390,25 @@ def plot_bands(arr, title=None, cmap="Greys_r",
             else:
                 ax.set(title='Band %i' %band)
             ax.set(xticks=[], yticks=[])
-        # This loop clears out the plots for bands 8-9 which are empty
-        # These plots are req by matplotlib when you specify plot rows & cols
+        # This loop clears out the plots for axes which are empty
+        # A matplotlib grid is always x by x. eg: 8 bands with 3 cols
         for ax in axs_ravel[total_layers:]:
             ax.set_axis_off()
             ax.set(xticks=[], yticks=[])
-
         plt.tight_layout()
-        return fig, axs
+        return fig
+
     elif arr.ndim == 2 or arr.shape[0] == 1:
         # If it's a 2 dimensional array with a 3rd dimension
-        if arr.shape[0] == 1:
-            arr = arr[0]
-        # Plot all bands
+        np.squeeze(arr)
+        # Plot one bands
         fig, ax = plt.subplots(figsize=figsize)
         ax.imshow(bytescale(arr), cmap=cmap,
                   extent=extent)
         if title:
             ax.set(title=title)
         ax.set(xticks=[], yticks=[])
-        return fig, ax
+        return fig
 
 
 def plot_rgb(arr, rgb=(0, 1, 2),
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -355,13 +355,20 @@ def plot_bands(arr, title=None, cmap="Greys_r",
     ...              figsize=(12,5),
     ...              cols=2)
     """
+
+    try:
+        arr.ndims
+    except AssertionError:
+        raise AssertionError("Input arr should be a numpy array")
+
     # If the array is 3 dimensional setup grid plotting
     if arr.ndim > 2 and arr.shape[0] > 1:
         # Test if there are enough titles to create plots
         if title:
             if not (len(title) == arr.shape[0]):
-                raise ValueError("""The number of plot titles should equal 
+                raise ValueError("""The number of plot titles should equal
                                  the number of array raster layers.""")
+
         # Calculate the total rows that will be required to plot each band
         plot_rows = int(np.ceil(arr.shape[0] / cols))
         total_layers = arr.shape[0]
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -392,7 +392,8 @@ def plot_bands(arr, title=None, cmap="Greys_r",
                 ax.set(title='Band %i' %band)
             ax.set(xticks=[], yticks=[])
         # This loop clears out the plots for axes which are empty
-        # A matplotlib grid is always x by x. eg: 8 bands with 3 cols
+        # A matplotlib axis grid is always uniform with x cols and x rows
+        # eg: an 8 band plot with 3 cols will always be 3 x 3
         for ax in axs_ravel[total_layers:]:
             ax.set_axis_off()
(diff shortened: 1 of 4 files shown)


lwasser:
@mbjoseph  will you please have a look at this? i think it should be 100% coverage but codecov is doing a few odd things.  note that i haven't pulled plotting out of spatial yet. i'm a bit torn

if i pull it from spatial it becomes more complex to remember where all of the tools are... in what module. then again that is how other python libraries work. the plot tests are all together however in a plot_test file.

lwasser:
this was working just the other day and now those tests are failing and i'm not sure why. it all works locally.

lwasser:
ok specifically if you look here it says these lines aren't covered. but i wrote tests for a single band array so they should be covered. i'm not sure why codecov things they are not covered.

https://codecov.io/gh/earthlab/earthpy/pull/122/src/earthpy/spatial.py#L401

lwasser:
i see that it's showing that it doesn't run because the coverage is red. .. but am not sure why that is given i can make it work on my machine.

mbjoseph:
I'm working locally for now. Have you tried modifying the `earthpy/spatial.py` `plot_bands` function definition to raise an error in that code block, then running `pytest` to verify that the tests fail?

lwasser:
sorry how do i run the tests locally? i actually am just manually running the code

lwasser:
but yes when i add that value error like you did i can raise it - WHEN i actually run the contents of  my test

mbjoseph:
To run tests locally you can run `pytest` in the terminal: https://github.com/earthlab/earthpy#testing

lwasser:
totally missed that. weird.  what i get is below.
Wonder if i need to do an envt refresh or something? if you have seen this before let me know. but might sign off for the night as.. not sure i'm ready to tackle envt issues just now :) 
```
============================================================ 13 passed in 2.44 seconds ============================================================
Exception ignored in: <bound method Handle.__del__ of <rtree.index.PropertyHandle object at 0x82432db70>>
Traceback (most recent call last):
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 875, in __del__
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 864, in destroy
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/core.py", line 35, in check_void_done
AttributeError: 'NoneType' object has no attribute 'Error_GetErrorCount'
Exception ignored in: <bound method Handle.__del__ of <rtree.index.IndexStreamHandle object at 0x82432da20>>
Traceback (most recent call last):
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 875, in __del__
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 864, in destroy
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/core.py", line 35, in check_void_done
AttributeError: 'NoneType' object has no attribute 'Error_GetErrorCount'
Exception ignored in: <bound method Handle.__del__ of <rtree.index.PropertyHandle object at 0x1073ae4e0>>
Traceback (most recent call last):
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 875, in __del__
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 864, in destroy
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/core.py", line 35, in check_void_done
AttributeError: 'NoneType' object has no attribute 'Error_GetErrorCount'
Exception ignored in: <bound method Handle.__del__ of <rtree.index.IndexStreamHandle object at 0x824357c18>>
Traceback (most recent call last):
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 875, in __del__
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/index.py", line 864, in destroy
  File "/Users/leah-su/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/rtree/core.py", line 35, in check_void_done
AttributeError: 'NoneType' object has no attribute 'Error_GetErrorCount'
```

mbjoseph:
Ahhh @lwasser I think I figured it out. Pytest discovers tests by looking for specific things (e.g., functions named `test_*`: https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery

The functions that would execute that `elif` block are probably not being discovered by pytest, because the function names are not prefixed by `test_`. If you change the names to `test_single_band_3dims` and `test_single_band_2dims`, that should cause that code to execute.

lwasser:
OH!! omg i didnt realize that. i'll do that now. thank you

lwasser:
THANK YOU SO MUCH @mbjoseph i would NOT have guessed that naming mattered. It totally worked and threw my test errors and such!! just pushed - let's see what codecov says now :)  this has been driving me a bit crazy as the code all seemed correct. i appreciate it.

lwasser:
OK @mbjoseph @joemcglinchy - can one of you give this puppy a review? I feel good about it now. :) 
And i've learned a LOT through this pr.

I"m going to open an issue about this plot function as it doesn't plot quietly unless you assign the call to an output var to store the fig / ax objects. If you return plt.show() then it just plots with no plot object which is nice but also is a bit limiting. Just a note. but i'd like to merge this for the time  being and think more about bigger picture plotting in another PR after a review.

earthpy/spatial.py
```diff
@@ -346,22 +347,34 @@ def plot_bands(arr, title=None, cmap="Greys_r",
     --------
     >>>import earthpy.spatial as es
     ...
```
mbjoseph: Do we also need `import numpy as np` in this example?

earthpy/spatial.py
```diff
@@ -346,22 +347,34 @@ def plot_bands(arr, title=None, cmap="Greys_r",
...
+
+    try:
+        arr.ndim
+    except AttributeError:
+        "Input arr should be a numpy array"
+
+
+    if title:
+        if (arr.ndim == 2) and (len(title) > 1):
+            raise ValueError("""You have provided more than one title for a
```
mbjoseph: May be worth being super explicit about what is expected here, e.g., "N titles were provided for a single band array, but plot_bands() expects exactly one title for a single band array", pasting in the number of titles that were provided for "N".

earthpy/spatial.py
```diff
@@ -377,19 +390,18 @@ def plot_bands(arr, title=None, cmap="Greys_r",
             else:
                 ax.set(title='Band %i' %band)
             ax.set(xticks=[], yticks=[])
-        # This loop clears out the plots for bands 8-9 which are empty
-        # These plots are req by matplotlib when you specify plot rows & cols
+        # This loop clears out the plots for axes which are empty
+        # A matplotlib grid is always x by x. eg: 8 bands with 3 cols
```
mbjoseph: Not sure I understand this comment - "x by x" would lead me to think that matplotlib grids are always square (x rows, x columns)

earthpy/tests/test_plot.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+""" General functions for matplotlib elements """
+
+
+# Create tuple
+tuups = (1, 2)
+
+im = np.random.randint(10, size=(2, 4, 5))
+single_band = im[0]
```
mbjoseph: The `single_band` object seems to only be used by one test (`test_num_titles`) - would it make sense to define it within that function, rather than in the global environment?

earthpy/tests/test_plot.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_arr_parameter():
+    """Error raised is not provided with an array"""
```
mbjoseph: The docstring for this function took me a few reads to understand. How about something like "If an array is not provided, raise an AttributeError"

earthpy/spatial.py
```diff
@@ -319,19 +319,20 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
-       quick visualization.
+    """Plot each layer in a raster stack read from rasterio in
+    (band, row ,col) order as a numpy array. plot_bands will create an
+    individual plot for each band in a grid.
 
     Parameters
     ----------
     arr: numpy array
-        An n-dimensional numpy array
+        An n-dimensional numpy array in (band, row, col) order
```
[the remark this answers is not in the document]
lwasser: i think i'd like to keep this as is a band, row, col as a convention for raster data. and i think rasterio too. n confuses the vars a bit more.

earthpy/spatial.py
lwasser: fixed!!

earthpy/spatial.py
lwasser: how do i write this... matplotlib will always create an even grid. not square  but...

if you have 8 bands, 8 plots and specify 3 cols, it will produce 9 axes - the last will be blank. 8 bands, 8 plots, 2 cols ... 8 axes will be produced exactly.

Does that make sense?

earthpy/tests/test_plot.py
lwasser: fixed!!

earthpy/spatial.py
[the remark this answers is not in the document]
mbjoseph: Cool - would you be comfortable with changing "in (band, row, col) order" to "with shape (band, row, col)", to better match numpy array terminology? I'm also thinking this might avoid confusing users, because "order" is also used to talk about the order of 2d arrays (e.g., row vs. column major order).

earthpy/spatial.py
lwasser: added!!

lwasser:
ok... i think this is ready for another review @mbjoseph  !!! still getting a warning running pytest but the tests pass..

earthpy/spatial.py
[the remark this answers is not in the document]
mbjoseph: Rasterio uses "shape" instead of "order" on that line, so wouldn't changing to "with shape (band, row, col)" be more consistent with rasterio than using "order"? I'm going to update my suggested change, since I'm not sure I was super clear in my last comment

earthpy/spatial.py
```diff
@@ -319,19 +319,20 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
-       quick visualization.
+    """Plot each layer in a raster stack read from rasterio in
+    (band, row , col) order as a numpy array. plot_bands will create an
+    individual plot for each band in a grid.
 
     Parameters
     ----------
     arr: numpy array
-        An n-dimensional numpy array
+        An n-dimensional numpy array with the order and numpy shape: (nbands, nrows, ncols)
```
lwasser: how about this as a middle ground @mbjoseph  ?

lwasser:
@mbjoseph  i think you've accepted all of the changes but the review is still not "approved" do you see an option to approve this pr? Or should i just override the review since we've don't it and merge as admin?

lwasser:
HAHAHAHA!  best giffy ever

PR #126, opened 2018-11-14
added 2 tests to hopefully get full coverage of mask.py

joemcglinchy:


earthpy/tests/test_mask.py
```diff
@@ -28,16 +28,26 @@ def test_masked_arr_returned():
...
+        how about a singular value? if not that, maybe a dictionary?"""
+
+    with pytest.raises(AttributeError):
+        make_apply_mask(im,
+                    mask_arr=im_mask,
+                    vals=0)
 
 # Test that a mask is provided to apply mask that doesn't have 1's in it.
 # should return - there's nothing to mask here!
+def test_mask_provided_with_no_ones():
```
lwasser: Can you kindly turn the comments into docstrings that describe what the test does @joemcglinchy  ?? this is really great!!

earthpy/tests/test_mask.py
```diff
@@ -28,16 +28,26 @@ def test_masked_arr_returned():
...
+
+    im_mask_no_ones = im_mask*0
+    with pytest.raises(ValueError):
+        make_apply_mask(im, im_mask_no_ones, [3,4])
 
 # test that a user provides an array with a mask already. does it merge properly?
-# and return a masked array with the values expected?
\ No newline at end of file
+# and return a masked array with the values expected?
+def test_im_with_mask_as_masked_array():
```
lwasser: i might be missing something here so please say the word if i am... where is the actual test in this function @joemcglinchy  ? i don't see any sort of assert or raises statement here but again might be missing something.

earthpy/tests/test_mask.py
```diff
@@ -28,16 +28,26 @@ def test_masked_arr_returned():
...
 
-# def test_val_list_provided:
-#     """Tuple works just fine. what about some other object type? like? """
-#
-#     with pytest.raises(AttributeError):
-#         make_apply_mask(im,
-#                     mask_arr=im_mask,
-#                     vals=[0,4])
+def test_val_list_provided():
+    """Tuple works just fine. what about some other object type? like?
```
lwasser: This is great. does a single value raise an assert? i can't quite see as it doesn't seem like travis ran or something but this is a good idea... i didn't realize a tuple would pass but didn't think to try a single val!

earthpy/tests/test_mask.py
joemcglinchy: yes, because the section of code that was not previously hit does not raise an error... it only checks if there is not a single value 1 element in the mask array.

earthpy/tests/test_mask.py
joemcglinchy: a single value should raise an assert, yes. tuples and lists both have a sort method, so a single value, or dictionary, etc.,  should do the trick.)

earthpy/tests/test_mask.py
joemcglinchy: I wasn't sure how to test for this one... we could try an assert statement to make sure the output is of type `np.ma.MaskedArray`... would that suffice?

lwasser: totally @joemcglinchy  

here is an example using hasattr
https://github.com/earthlab/earthpy/blob/master/earthpy/tests/test_clip.py#L67

I think you can do something similar using
assert `np.ma.is_masked()`

https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.ma.is_masked.html

does that seem reasonable to you?

lwasser:
@joemcglinchy it looks like we have full coverage now on mask. If you can make the few small changes (turning the comments into doc strings) i'll merge this and then i think we can potentially merge this all together into earthpy!! Thank you again.

earthpy/tests/test_mask.py
joemcglinchy: added to it in the most recent commit. push away!

lwasser: woo hoo!! thank you @joemcglinchy  i'll merge this now.

PR #127, opened 2018-11-16
first round of tests for bytescale

joemcglinchy:
added first round of tests for bytescale. still need a test for the case where cmin=cmax, but that's an odd one at first glance.

### DIFF
--- a/earthpy/tests/test_spatial.py
+++ b/earthpy/tests/test_spatial.py
@@ -1,6 +1,45 @@
-from earthpy import spatial
+from earthpy import spatial as es
+import numpy as np
+import pytest
 
 
-def test_dummy():
-    """"Dummy test that will pass."""
-    assert 1 == 1
+def test_bytescale_high_low_val():
+    """"Unit tests for earthpy.spatial.bytescale """
+    
+    arr = np.random.randint(300, size=(10,10))
+    
+    # Bad high value
+    with pytest.raises(ValueError):
+        es.bytescale(arr, high=300)
+        
+    # Bad low value
+    with pytest.raises(ValueError):
+        es.bytescale(arr, low=-100)
+        
+    # High value is less than low value
+    with pytest.raises(ValueError):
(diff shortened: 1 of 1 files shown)


lwasser:
hey there @joemcglinchy can you kindly update your fork / branch? there is a conflict here and i think  it's involved with some other tests for other functions that were already on master that are not reflected in your PR. let me know if that is clear as mud :)

joemcglinchy:
not totally clear... update it with what exactly? a `git pull` from an `upstream` branch?

lwasser:
yes you need to update your fork and branch from earthpy/master. YOu can do a reverse pull request to update which is very straight forward! but there are conflicts because you need to update things! drop by if you want me to show you!! git pull from upstream should also work BUT it might be weird as you might need to rebase too?? a reverse pull request should do the trick. i can create one to show you... one min.

lwasser:
https://github.com/joemcglinchy/earthpy/pull/1

see the above PR to see how many commits behind you are!!

joemcglinchy:
makes sense since the `bytescale_tests` was a branch of the `master` branch from my fork, which hadn't been updated!

lwasser:
oops @joemcglinchy you merged this into master.

joemcglinchy:
huh. that's weird! thought I merged it into `bytescale_tests`! at least, that's what I thought I saw here: https://github.com/joemcglinchy/earthpy/pull/1#issuecomment-440043239

lwasser:
is it possible that you pushed up to upstream instead of origin? Did you do anything local? If you did i suspect you pushed to upstream rather than your fork origin and that caused the merge to happen. that would make sense. Maybe you could consider removing the upstream remote if that becomes and issue just to avoid it happening. 

I think this because i see your master branch on your fork is still not up to date with earthpy master -- [see here](https://github.com/joemcglinchy/earthpy/pull/2)  

if you merged, pulled things down locally and pushed to upstream rather than your fork it would merge this PR.

Issue #118, opened 2018-11-08
Move generation test functions -- to utils.py script


lwasser:
@mbjoseph mentioned he is willing to create a test_utils script (mayeb diff name) that contains util functions that we will use throughout to do things like
1. create geodataframes
2. create tif files and maybe??
3. numpy arrays

just so they are all in the same place .
@mbjoseph if i misheard this in our meeting please unassign yourself!

PR #128, opened 2018-11-26
Adding pytest fixtures for re-used test data

mbjoseph:
Our original idea was to add a utility functions script to the tests directory, and then use those utility functions in test modules to simulate fake data. But the best practice for pytest seems instead to use fixtures, which are functions that generate objects required for tests to run. Each test takes fixture names as arguments, and then executes those fixture functions to create fake data independent for each test. Creating the data independently for each test is more robust than creating shared data in a global environment that is used by multiple tests. 

Fixtures used by multiple test modules are defined in `earthpy/tests/conftest.py`. 

I found these links on fixtures to be helpful in wrapping my head around the concept: 

- https://docs.pytest.org/en/latest/fixture.html
- https://blog.daftcode.pl/the-cleaning-hand-of-pytest-28f434f4b684

@lwasser have a look, and let me know if you have questions/changes! I added some info to the README to remind ourselves later of how we use fixtures. I also cleaned up some flake8 and pylint formatting issues in some of the other test modules as part of this PR. Relates to #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ This package uses [pytest](https://pytest.org/) for tests.
 To run tests locally, execute the command `pytest` from the command line:
 
 ```
-pytest
+>>>pytest
 ```
 
 ### Fake data for testing
--- a/README.md
+++ b/README.md
@@ -14,11 +14,15 @@ To install, use pip. `--upgrade` is optional but it ensures that the package ove
 when you install and you have the current version. If you don't have the package
 yet you can still use the `--upgrade` argument.
 
-`pip install --upgrade git+https://github.com/earthlab/earthpy.git`
+```bash
+$ pip install --upgrade git+https://github.com/earthlab/earthpy.git
+```
 
 Then import it into python.
 
-`import earthpy as et`
+```python
+>>> import earthpy as et
+```
 
 
 ## Contributors
@@ -33,14 +37,14 @@ Then import it into python.
 This package uses [pytest](https://pytest.org/) for tests.
 To run tests locally, execute the command `pytest` from the command line:
 
-```
->>>pytest
+```bash
+$ pytest
 ```
 
-### Fake data for testing
+### Data generated for testing
 
 If a test requires a data object such as a GeoDataFrame or numpy array, and
-copies of that data object is required by multiple tests, we can use [pytest
+copies of that data object are required by multiple tests, we can use [pytest
 fixtures](https://docs.pytest.org/en/latest/fixture.html) to cleanly create
 and tear down those objects independently for each test.
 
--- a/README.md
+++ b/README.md
@@ -31,4 +31,19 @@ Then import it into python.
 ## Testing
 
 This package uses [pytest](https://pytest.org/) for tests.
-To run tests locally, execute the command `pytest` from the command line.
+To run tests locally, execute the command `pytest` from the command line:
+
+```
+pytest
+```
+
+### Fake data for testing
+
+If a test requires a data object such as a GeoDataFrame or numpy array, and
+copies of that data object is required by multiple tests, we can use [pytest
+fixtures](https://docs.pytest.org/en/latest/fixture.html) to cleanly create
+and tear down those objects independently for each test.
+
+See [`earthpy/tests/conftest.py`](earthpy/tests/conftest.py) for fixture
+definitions, and [`earthpy/tests/test_clip.py`](earthpy/tests/test_clip.py)
+for example usage of fixtures in tests.
--- a/earthpy/tests/conftest.py
+++ b/earthpy/tests/conftest.py
@@ -1,0 +1,91 @@
+""" Utility functions for tests. """
+import numpy as np
+import pytest
+from shapely.geometry import Polygon, Point, LineString
+import geopandas as gpd
+
+
+def make_locs_gdf():
+    """ Create a dummy point GeoDataFrame. """
+    pts = np.array([[2, 2],
+                    [3, 4],
+                    [9, 8],
+                    [-12, -15]])
+    gdf = gpd.GeoDataFrame([Point(xy) for xy in pts],
+                           columns=['geometry'],
+                           crs={'init': 'epsg:4326'})
+    return gdf
+
+
+def make_poly_in_gdf():
+    """ Bounding box polygon. """
+    poly_inters = Polygon([(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)])
+    gdf = gpd.GeoDataFrame([1],
+                           geometry=[poly_inters],
+                           crs={'init': 'epsg:4326'})
+    return gdf
+
+
+def make_locs_buff():
+    """ Buffer points to create multi poly. """
+    buffered_locations = make_locs_gdf()
+    buffered_locations["geometry"] = buffered_locations.buffer(4)
+    return buffered_locations
+
+
+def make_donut_geom():
+    """ Make a donut geometry. """
+    donut = gpd.overlay(make_locs_buff(), make_poly_in_gdf(),
+                        how='symmetric_difference')
+    return donut
+
+
+@pytest.fixture
+def locs_gdf():
+    """ Get a dummy point GeoDataFrame.
+
+    This fixture calls make_locs_gdf(), which is a function that is used in
+    multiple fixtures. But, fixtures are not supposed to be used like that:
+
+    see https://github.com/pytest-dev/pytest/issues/3950 for discussion
+    """
+    return make_locs_gdf()
+
+
+@pytest.fixture
+def linez_gdf():
+    """ Create Line Objects For Testing """
+    linea = LineString([(1, 1), (2, 2), (3, 2), (5, 3)])
+    lineb = LineString([(3, 4), (5, 7), (12, 2), (10, 5), (9, 7.5)])
+    gdf = gpd.GeoDataFrame([1, 2],
+                           geometry=[linea, lineb],
+                           crs={'init': 'epsg:4326'})
+    return gdf
+
+
+@pytest.fixture
+def poly_in_gdf():
+    """ Fixture for a bounding box polygon. """
+    return make_poly_in_gdf()
+
+
+@pytest.fixture
+def locs_buff():
+    """ Fixture for buffered locations. """
+    return make_locs_buff()
+
+
+@pytest.fixture
+def donut_geom():
+    """ Fixture for donut geometry objects. """
+    return make_donut_geom()
+
+
+@pytest.fixture
+def multi_gdf():
+    """ Create a multi-polygon GeoDataFrame. """
+    multi_poly = make_donut_geom().unary_union
+    out_df = gpd.GeoDataFrame(gpd.GeoSeries(multi_poly),
+                              crs={'init': 'epsg:4326'})
+    out_df = out_df.rename(columns={0: 'geometry'}).set_geometry('geometry')
+    return out_df
--- a/earthpy/tests/test_clip.py
+++ b/earthpy/tests/test_clip.py
@@ -1,70 +1,13 @@
 """Tests for the clip module."""
 
-import numpy as np
 import pytest
-from shapely.geometry import Polygon, Point, LineString
+from shapely.geometry import Polygon
 import shapely
 import geopandas as gpd
 import earthpy.clip as cl
 
 
-""" Setup dummy data for tests """
-
-pts = np.array([[2, 2],
-                [3, 4],
-                [9, 8],
-                [-12, -15]])
-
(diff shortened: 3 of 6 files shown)


mbjoseph:
I am going to look into this reduction in coverage, so hold off on review for now please!

mbjoseph:
Ok @lwasser this is ready for review now!

mbjoseph:
@lwasser I just noticed there are now some conflicting files for this PR. I will make changes to resolve these. Maybe we can discuss linting here: https://github.com/earthlab/earthpy/issues/132

earthpy/tests/test_spatial.py
```diff
@@ -48,44 +41,39 @@ def test_extent_to_json():
...
     with pytest.raises(ValueError):
         es.bytescale(arr, cmin=100, cmax=50)
-        
-        
+
     # Test scale value max is less equal to min
     es.bytescale(arr, cmin=100, cmax=100)
-    # assert something    
-        
+    # TODO: add assertion here
```
lwasser: Yes... so we had a bytescale snafu today. and Joe commented this out completely but then we realized it wasn't actually the cause of the issue at all.. so you're code here might be just fine. I suspect travis was going wonky.

mbjoseph: OK - I commented out this line to go with the most recent version, and left the TODO part to remind us to write a real test for this case.

mbjoseph:
Ok @lwasser I fixed the merge conflict - this should be ready for a review now!

lwasser:
weird. i still see two conflicts here @mbjoseph i merged clip just today. perhaps you can fix these? i'll go ahead with a review regardless.

README.md
```diff
@@ -31,4 +31,19 @@ Then import it into python.
 ## Testing
 
 This package uses [pytest](https://pytest.org/) for tests.
-To run tests locally, execute the command `pytest` from the command line.
+To run tests locally, execute the command `pytest` from the command line:
+
+```
+pytest
```
lwasser: Why don't we use

```
>>>pytest
```
seems to be a convention for CLI statements. We can do this everywhere too.. all of my install instructions should be updated following this!

see
https://github.com/mapbox/rasterio/blob/master/README.rst
for an example

README.md
```diff
@@ -31,4 +31,19 @@ Then import it into python.
 ## Testing
 
 This package uses [pytest](https://pytest.org/) for tests.
-To run tests locally, execute the command `pytest` from the command line.
+To run tests locally, execute the command `pytest` from the command line:
+
+```
+>>>pytest
```
lwasser: pls lemme know if this didn't work as expected - ie added a small commit here?

README.md
```diff
@@ -31,4 +31,19 @@ Then import it into python.
...
+```
+
+### Fake data for testing
```
lwasser: what about a word that is nicer than Fake? `### Data generated for testing`

README.md
```diff
@@ -31,4 +31,19 @@ Then import it into python.
...
+
+If a test requires a data object such as a GeoDataFrame or numpy array, and
+copies of that data object is required by multiple tests, we can use [pytest
```
lwasser: copies of that data are required

earthpy/tests/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def make_donut_geom():
+    """ Make a donut geometry. """
+    donut = gpd.overlay(make_locs_buff(), make_poly_in_gdf(),
+                        how='symmetric_difference')
+    return donut
+
+
+@pytest.fixture
```
lwasser: this is neat. what does this @pytest.fixture call actually do for this function? i'm more used to this in class definitions. and once this is setup how do we call these data in the tests?

earthpy/tests/test_clip.py
```diff
@@ -1,96 +1,46 @@
...
-        cl.clip_shp(locs_gdf, poly_out_gdf)
+        cl.clip_shp(unit_gdf, non_overlapping_gdf)
 
 
-def check_input_gdfs():
+def check_input_gdfs(poly_in_gdf):
     """Test that function fails if not provided with 2 GDFs."""
     with pytest.raises(AssertionError):
-        cl.clip_shp(locs, poly_in_gdf)
+        cl.clip_shp([], poly_in_gdf)
```
lwasser: just curious. i know this is passing an empty list. is this good form over calling `list()` ? it seems like a mistake when i look at it but i know what you are doing. would `list()` be more explicit?

earthpy/tests/test_clip.py
```diff
@@ -1,96 +1,46 @@
...
-def test_returns_gdf():
+def test_returns_gdf(locs_gdf, poly_in_gdf):
     """Test that function returns a GeoDataFrame (or GDF-like) object."""
     out = cl.clip_shp(locs_gdf, poly_in_gdf)
     assert hasattr(out, 'geometry')
 
 
 def test_non_overlapping_geoms():
     """Test that a bounding box returns error if the extents don't overlap"""
+    unit_box = Polygon([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)])
```
lwasser: another question - dont fixtures replace creating data in line? i may just not understand

earthpy/tests/test_clip.py
```diff
@@ -1,96 +1,46 @@
...
 
-def check_input_gdfs():
+def check_input_gdfs(poly_in_gdf):
     """Test that function fails if not provided with 2 GDFs."""
     with pytest.raises(AssertionError):
-        cl.clip_shp(locs, poly_in_gdf)
+        cl.clip_shp([], poly_in_gdf)
     with pytest.raises(AssertionError):
-        cl.clip_shp(poly_in_gdf, locs)
+        cl.clip_shp(poly_in_gdf, [])
```
lwasser: yea let's use list() instead. just keep seeing this `[]` and thinking it's a mistake

earthpy/tests/test_clip.py
```diff
@@ -100,19 +50,19 @@ def test_clip_poly():
 # TODO same test for points and lines
```
lwasser: oopsy. i believe i did this TODO??

earthpy/tests/test_clip.py
```diff
@@ -100,19 +50,19 @@ def test_clip_poly():
...
 
 
-def test_clip_multipoly():
+def test_clip_multipoly(poly_in_gdf, multi_gdf):
```
lwasser: here you are calling the fixtures. is that right?

earthpy/tests/test_clip.py
```diff
@@ -1,96 +1,46 @@
 """Tests for the clip module."""
 
-import numpy as np
 import pytest
```
lwasser: how are the fixtures imported here? ie how does it know where to find those functions if they are not imported? magical?

earthpy/tests/test_mask.py
```diff
@@ -1,9 +1,9 @@
+""" Tests for the mask module. """
```
lwasser: much better up here!

earthpy/tests/test_spatial.py
```diff
@@ -1,19 +1,12 @@
...
 import pandas as pd
-import numpy as np
 import pytest
-from shapely.geometry import Polygon, Point, LineString
-import shapely
+from shapely.geometry import Polygon, Point
 import geopandas as gpd
-import earthpy.spatial as es
-import earthpy.clip as cl
+from earthpy import spatial as es
```
lwasser: i don't normally import this way. why did you adjust this one instance of the import? 

alternatively if we want to be more explicit... could say:

`from earthpy.spatial import extent_to_json, bytescale`

earthpy/tests/test_spatial.py
```diff
@@ -48,44 +41,39 @@ def test_extent_to_json():
     with pytest.raises(AssertionError):
```
lwasser: if we do the above suggestion this would then just be `extent_to_json` (without the `es.`)

mbjoseph:
Hi @lwasser - looks like these are new conflicts. I will fix these too then make changes as per your review!

README.md
mbjoseph: I think rasterio is using `$` to indicate bash commands, and `>>>` to indicate Python commands. So, to be consistent with them maybe I will change this to 

```bash
$ pytest
```

lwasser:
Ahhh you’re right … maybe `$` then? I never looked that closely! R has one > and python >>>

I think the dollar sign is better for bash looking at other readme’s!!

earthpy/tests/conftest.py
mbjoseph: By adding `@pytest.fixture` as a decorator for a function in `conftest.py`, we are making the `locs_gdf` function available for all `earthpy/tests/test_*.py` modules. This is a way to generate data that are shared by multiple tests that span different test modules. See here for more: https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

mbjoseph: Examples of calling these fixtures in tests can be found in `test_clip.py`, where they show up as function arguments.

earthpy/tests/test_clip.py
mbjoseph: Fixtures are good for defining data that are shared by multiple tests. So, if we wanted to use this `unit_box` in multiple tests, we could create it as a fixture. But, currently it is only used in this test, so adding it as a fixture might add some unnecessary complexity.

earthpy/tests/test_clip.py
mbjoseph: Yep - I think that was one of yours in this commit: https://github.com/earthlab/earthpy/commit/77e2ecd28509d138f50a20e51d0dffc2e278fc6d#diff-6c8697d6bac09c332f11fca7ec8cf5ffR87

earthpy/tests/test_clip.py
mbjoseph: Yes! They act as function arguments, so that you can explicitly see what data are needed for each test.

README.md
mbjoseph: Ok I have changed all bash commands to use $ and all python commands to use >>>

README.md
mbjoseph: Fixed this too

README.md
mbjoseph: Good idea - made this change!

README.md
mbjoseph: Thanks!

earthpy/tests/test_clip.py
mbjoseph: Changed to `list()`

earthpy/tests/test_clip.py
mbjoseph: Sure! changed to `list()`

earthpy/tests/test_spatial.py
mbjoseph: Ooops. Changed to `import earthpy.spatial as es`. I like to use `es.some_fun()` over importing particular functions to be super clear in the code where functions are coming from whenever they are called.

lwasser: yaa... i think best python practice is to keep the namespace clean by just importing the functions you want. have you gotten to that part in our OOP book yet? The explicit part comes via the imports! for the spatial module i normally do import  `import earthpy.spatial as es` because often times i use so many functions from it. But like for the show() function in rasterio i just import that fun at the top OR the glob library i normally just import glob. looking at rasterio they are pretty explicit with just importing functions. potential danger zone is two packages, same fun name. i like what you have above but might suggest we change it in the future (or not!) after chatting with a few more folks about namespace best practices :)

README.md
```diff
@@ -31,4 +35,19 @@ Then import it into python.
 ## Testing
 
 This package uses [pytest](https://pytest.org/) for tests.
-To run tests locally, execute the command `pytest` from the command line.
+To run tests locally, execute the command `pytest` from the command line:
+
+```bash
+$ pytest
```
lwasser: this looks great!! then it's super clear where to run the command. i like it.

earthpy/tests/conftest.py
lwasser: perfect. ok i get it. thanks!!

earthpy/tests/test_clip.py
lwasser: got it.

earthpy/tests/test_clip.py
lwasser: neat. this is a cool way to structure tests. i may have to redo some of the other tests for plotting based upon this now.

README.md
lwasser: woohoo!!

lwasser:
This looks great! merging. thank you @mbjoseph !!

PR #135, opened 2018-11-30
[tests] add non gdf tests to clip test file 100% coverage

lwasser:
This should magically fill the test cov to 100% coverage. added 2 non gdf tests

### DIFF
--- a/earthpy/clip.py
+++ b/earthpy/clip.py
@@ -112,14 +112,12 @@ def clip_shp(shp, clip_obj):
     """
     try:
         shp.geometry
-    except AttributeError:
-        raise AttributeError("""Please provide a geoDataFrame with a valid
-                             geometry column""")
-    try:
         clip_obj.geometry
     except AttributeError:
-        raise AttributeError("""Please provide a geoDataFrame with a valid
+        raise AttributeError("""Please make sure that your input and clip
+                             GeoDataFrames have a valid
                              geometry column""")
+
 
     if not any(shp.intersects(clip_obj)):
         raise ValueError("Shape and crop extent do not overlap.")
--- a/earthpy/clip.py
+++ b/earthpy/clip.py
@@ -112,13 +112,13 @@ def clip_shp(shp, clip_obj):
     """
     try:
         shp.geometry
-    except AssertionError:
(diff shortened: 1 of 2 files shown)


lwasser:
or joe could look too. just adding full coverage to the clip module. seemed like low hanging fruit.

lwasser:
Same here. Ready for a review

earthpy/clip.py
```diff
@@ -112,13 +112,13 @@ def clip_shp(shp, clip_obj):
     """
     try:
         shp.geometry
-    except AssertionError:
-        raise AssertionError("""Please provide a geoDataFrame with a valid
+    except AttributeError:
```
mbjoseph: These two try/except calls have a duplicate `AttributeError`. This could be DRY-ified as follows:

```python
try:
    shp.geometry
    clip_obj.geometry
except AttributeError:
    raise AttributeError("...")
```

Or, if you wanted to use different error messages to tell users explicitly that the `shp` or `clip_obj` input is invalid:

```python
try:
    shp.geometry
except AttributeError:
    # tell users the shp object needs a geometry column

try:
    clip_obj.geometry
except AttributeError:
    # tell users that clip_obj needs a geometry column
```

Also should `geoDataFrame` be `GeoDataFrame`?

earthpy/tests/test_clip.py
```diff
@@ -63,6 +63,14 @@
 
 """ Run clip shape tests """
 
+def test_not_gdf():
+    """Test that an error is raised when provided with a non geodataframe
```
mbjoseph: Minor grammar/spelling check for this docstring -- spelling for GeoDataFrame, indefinite article with plural noun ("a non geodataframe objects")

```suggestion
    """Non-GeoDataFrame inputs raise attribute errors."""
```

earthpy/clip.py
lwasser: Totally. I"m going to combine them. I think the intent originally was to have custom messages but looking at it now...it's just as clear to combine.

lwasser:
ok @mbjoseph  changes made! lemme know if you catch something else OR if we can merge this bad boy.

PR #138, opened 2018-12-04
List package versions in Travis

mbjoseph:
@lwasser This PR should force Travis to print the output from `conda list` after activating the environment. That way we can see what versions of each package are installed in Travis for debugging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
   - conda info -a
   - conda env create -f environment.yml
   - source activate earthpy-dev
+  - conda list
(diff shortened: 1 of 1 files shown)


mbjoseph:
@lwasser do we want to keep this change? It seems useful for debugging issues with builds on Travis.

lwasser:
yes @mbjoseph  !! let's keep it. i'm still baffled by matplotlib v 2.x however!! it's super useful to see what's installed. Wanna merge if you feel ready to?! or do you want me to?

PR #140, opened 2018-12-05
 Modify normalized_diff and add tests

### LINKED ISSUE #2: Normalized diff function - What order would a user expect the bands to be in?

### LINKED ISSUE #121: Add unit tests to norm_diff function
in followup to our meeting today, @jlpalomino please add unit tests to the norm diff function!

jlpalomino:
@lwasser These changes update the normalized_diff function in spatial.py to modify when results are returned as masked arrays and add three tests for this function to test_spatial.py. fixes #121, fixes #2


earthpy/tests/test_spatial.py
```diff
@@ -42,6 +42,47 @@ def test_extent_to_json():
...
+
+def test_normalized_diff_nomask():
+    """If result of normalized_diff does not include nan or inf values,
+    array is returned as unmasked."""
+
+    # Test data
+    b1_2d = np.array([[1, 2, 3, 4, 5], [14, 12, 13, 14, 17]])
+    b2_2d = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+
+    n_diff = normalized_diff(b2=b2_2d, b1=b1_2d)
```
lwasser: This is a great start @jlpalomino looks like it's failing some tests  :)

[see travis results]( https://travis-ci.org/earthlab/earthpy/builds/464008139#L1389 )
To begin you call normalized_diff here. This will work if you do 

 from earthpy.spatial import normalized_diff in the imports. But currently we have all of the spatial module loaded so the better fix for the time being is to just call

`es.normalized_diff()`

lwasser: I also suggest that given we are switching the default band order, that we use that order here. so just swap b1 and b2 parameter / argument order

earthpy/spatial.py
```diff
@@ -74,7 +74,15 @@ def normalized_diff(b1, b2):
...
 
     n_diff = (b2 - b1) / (b2 + b1)
-    n_diff = np.ma.masked_invalid(n_diff)
+
+    # Mask invalid values
+    if np.isnan(n_diff).any():
+        n_diff = np.ma.masked_invalid(n_diff)
+
+        # Set inf values to nan
+        if np.isinf(n_diff).any():
```
lwasser: This is a nice adjustment to ensure those cells have a na placeholder rather than inf. 
I think we should update the docstring to acknowledge that this is what the function does in the case of a calculation that doesn't have a finite answer.

earthpy/spatial.py
```diff
@@ -56,7 +56,7 @@ def normalized_diff(b1, b2):
     Returns
     ----------
     n_diff : ndarray with the same shape as inputs
-        The element-wise result of (b2-b1) / (b2+b1) with all nan values
+        The element-wise result of (b2-b1) / (b2+b1) with all nan and inf values
         masked.
```
lwasser: i think this is where we could have another statement here that says something like.:

 All values == Inf will  be set to np.nan.  (or something like that just to be explicit!

earthpy/tests/test_spatial.py
jlpalomino: These changes are now reflected in the tests for normalized_diff.

earthpy/spatial.py
jlpalomino: The docstring has been updated. Also note that I changed the order and hierarchy of operations, so that inf values are set to nan first, and the mask is not contingent on existing nan values. This change ensures that the array is masked any time there are nan values (whether the nan values were in the input arrays or whether the nan values are the result of the function).

earthpy/spatial.py
jlpalomino: The docstring has been updated to explicitly state that inf values are set to nan.

earthpy/spatial.py
```diff
@@ -44,20 +44,20 @@ def extent_to_json(ext_obj):
 # Calculate normalized difference between two arrays
 # Both arrays must be of the same size
 
-def normalized_diff(b1, b2):
-    """Take two numpy arrays and calculate the normalized difference
+def normalized_diff(b2, b1):
```
lwasser: This is great @jlpalomino . i think that perhaps my comment was unclear. 
did we decide to do something like
```
def normalized_diff(b1, b2):

Math will be calculated (b1-b2) / (b1+b2)

...
b1, b2 : arrays with the same shape

ndiff = es.normalized_diff(b1=nir_band, b2=red_band)
```

@joemcglinchy does this follow what you and i think jenny had +1 in the issue ??

What is the most intuitive to do here for a user. for me i'm wondering if we just want b1, b2 and then math is calculated b1-b2. but perhaps i'm odd in my thinking!

jlpalomino: @lwasser The band order has been updated to match your example of b1 and b2, per the group's decision.

jlpalomino:
@lwasser These changes update the band order in the normalized_diff function to b1 and b2 and add a custom warning for the division by zero to notify user that infinity values will be converted to nan values.

lwasser:
hey @jlpalomino !! this is looking great. Please do the following so we can approve and merge this
1. will you kindly double check that the `.rst` file documentation reflects that changes that you made here?
2. will you please update your fork from master and then fix all merge conficts? I see above there are a few now. I had the same thing with my PR's. This is because we've all been working on these files. Also we ran black on all files for formatting.

when this is all update to date i think we will be ready to squash and merge. Thank you!!

earthpy/tests/test_spatial.py
```diff
@@ -108,6 +109,63 @@ def test_extent_to_json():
...
+def test_normalized_diff_shapes():
+    """If provided with arrays of different shapes,
+    normalized_diff returns ValueError."""
+
+    # Test data
+    b1_2d = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    b2_1d = np.array([[1, 2, 3, 4, 5, 14, 12, 13, 14, 17]])
+
+    # Check ValueError
+    with pytest.raises(ValueError, match=r"Both arrays should be of the same dimensions"):
```
lwasser: What about saying `"Both arrays should have the same dimensions"`  both here and in the function err too?

earthpy/tests/test_spatial.py
```diff
@@ -108,6 +109,63 @@ def test_extent_to_json():
         es.extent_to_json([0, 1, 1, 0])
 
 
+def test_normalized_diff_shapes():
+    """If provided with arrays of different shapes,
+    normalized_diff returns ValueError."""
+
+    # Test data
+    b1_2d = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
```
lwasser: i think these could be fixtures
```
@pytest.fixture
def b1_b2_arrs():
    b1 = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
    b2 = np.array([[1, 2, 3, 4, 5], [14, 12, 13, 14, 17]])
    return b1, b2
```

function would then look like
```
def test_normalized_diff_shapes(b1_b2_arrs):
     """Test that two arrays with different shapes returns a ValueError."""

     # Test data
     b1, b2 = b1_b2_arrs
     b2 = b2[0] # or something like that to just select one part of the arr
 
     # Check ValueError
     with pytest.raises(ValueError, match=r"Both arrays should be of the same dimensions"):
         es.normalized_diff(b1 = b1, b2 = b2)
```

earthpy/tests/test_spatial.py
```diff
@@ -108,6 +109,63 @@ def test_extent_to_json():
...
+    b2_1d = np.array([[1, 2, 3, 4, 5, 14, 12, 13, 14, 17]])
+
+    # Check ValueError
+    with pytest.raises(ValueError, match=r"Both arrays should be of the same dimensions"):
+        es.normalized_diff(b1 = b1_2d, b2 = b2_1d)
+
+
+def test_normalized_diff_no_mask():
```
lwasser:
```
 def test_normalized_diff_no_mask(b1_b2_arrs):
     """If result of normalized_diff does not include nan values,
     array is returned as unmasked."""
 
     b1, b2 = b1_b2_arrs
     n_diff = es.normalized_diff(b1=b1, b2=b2)
 
     # Output array unmasked 
     assert ma.is_masked(n_diff) not True
```
Isn't it not the best idea to use `==` with booleans but rather just assert as assert will fail if it's false. The above should work :) (i think)

earthpy/tests/test_spatial.py
```diff
@@ -108,6 +109,63 @@ def test_extent_to_json():
...
+    b1_2d = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    b2_2d = np.array([[1, 2, 3, 4, 5], [14, 12, 13, 14, 17]])
+
+    n_diff = es.normalized_diff(b1 = b1_2d, b2 = b2_2d)
+
+    # Output array unmasked
+    assert ma.is_masked(n_diff) == False
+
+
+def test_normalized_diff_inf():
```
lwasser: Here i think you could use the same arrays just replace one value with -15

earthpy/tests/test_spatial.py
```diff
@@ -108,6 +109,63 @@ def test_extent_to_json():
...
+    # Output array masked
+    assert ma.is_masked(n_diff)
+
+
+def test_normalized_diff_mask():
+    """If result of normalized_diff includes nan values,
+    array is returned as masked."""
+
+    # Test data
+    b1_2d_nan = np.array([[6, 7, 8, 9, 10], [np.nan, 17, 18, 19, 20]])
```
lwasser: similarly here you could just add nans where needed using the same fixture.

lwasser:
ok @jlpalomino i left you some feedback. i'm still learning about fixtures so wouldn't mind a second opinion from @mbjoseph IF he has time. Max if i'm super off base please let me know. but it seemed like we could simplify this with just one fixture that you modify slightly when called to accommodate each unique test.

earthpy/tests/test_spatial.py
jlpalomino: The text for this ValueError has now been modified in the function and the test.

earthpy/tests/test_spatial.py
jlpalomino: Great point. I have replaced `== False` with `assert not`.

earthpy/tests/test_spatial.py
mbjoseph: I agree - using a fixture for these arrays will keep these tests nice and DRY :desert:

docs/spatial-raster.rst
```diff
@@ -89,26 +89,28 @@ calculates the normalized difference from them.
...
-      fill_value=1e+20)
+    >>> # Calculate normalized difference vegetation index
+    >>> nir_band = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    >>> red_band = np.array([[1, 2, 3, 4, 5], [11, 12, 13, 14, 15]])
+    >>> ndvi = es.normalized_diff(b1 = nir_band, b2 = red_band)
+
+    >>> # Calculate normalized burn ratio
+    >>> nir_band = np.array([[8, 10, 13, 17, 15], [18, 20, 22, 23, 25]])
+    >>> swir_band = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    >>> nbr = es.normalized_diff(b1 = nir_band, b2 = swir_band)
```
mbjoseph: Here we should print `ndvi` and `nbr`, and paste in the expected output so that doctest can verify that the output is what we expect. This adds a layer of testing to our documentation to make sure it stays current with the underlying code.

earthpy/spatial.py
```diff
@@ -47,44 +47,51 @@ def extent_to_json(ext_obj):
...
-             0.15151515151515152, 0.14285714285714285]],
-      mask=[[False, False, False, False, False],
-            [False, False, False, False, False]],
-      fill_value=1e+20)
+    ...
+    ... # Calculate normalized difference vegetation index
+    >>> nir_band = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    >>> red_band = np.array([[1, 2, 3, 4, 5], [11, 12, 13, 14, 15]])
+    >>> ndvi = es.normalized_diff(b1 = nir_band, b2 = red_band)
+    ...
```
mbjoseph: Also here we could have the expected output so that doctest can verify that the documentation is current with the code.

earthpy/spatial.py
```diff
@@ -47,44 +47,51 @@ def extent_to_json(ext_obj):
...
+    ... # Calculate normalized burn ratio
+    >>> nir_band = np.array([[8, 10, 13, 17, 15], [18, 20, 22, 23, 25]])
+    >>> swir_band = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    >>> nbr = es.normalized_diff(b1 = nir_band, b2 = swir_band)
```
mbjoseph: doctest output expectations could be provided here too

earthpy/tests/test_spatial.py
lwasser: how are you so good with emojis @mbjoseph  i don't even know how to find 🏜 !!

earthpy/tests/test_spatial.py
lwasser: yes after i wrote that i realized `assert not` is the simplest option!! great catch :)

docs/spatial-raster.rst
lwasser: ahhhh right. you know @mbjoseph i didn't fully understand the coolness of doctest UNTIL joe updated his functions and the docs failed. then i realized it is brilliant as another source of checks.

earthpy/tests/test_spatial.py
mbjoseph: This may be a silly question, but is the raw string literal necessary here for the value error message matching?

jlpalomino: @mbjoseph @joemcglinchy Not sure if there is a better way, but when I use `message = ` rather than `match=r`, I can change the text to not match, and the test still passes.

docs/spatial-raster.rst
jlpalomino: The output of printing ndvi and nbr have been added to normalized_diff in the .rst.

earthpy/spatial.py
jlpalomino: The output of printing ndvi and nbr have been added to normalized_diff in spatial.py.

earthpy/spatial.py
jlpalomino: The output of printing ndvi and nbr have been added to normalized_diff in spatial.py.

earthpy/tests/test_spatial.py
jlpalomino: Fixtures for these arrays have been implemented in testing of normalized_diff.

earthpy/tests/test_spatial.py
jlpalomino: Fixtures for these arrays have been implemented in testing of normalized_diff.

earthpy/tests/test_spatial.py
jlpalomino: Fixtures for these arrays have been implemented in testing of normalized_diff.

earthpy/tests/test_spatial.py
mbjoseph: Does the test work if you still use the `match` argument without the raw string literal? e.g., 

```python
with pytest.raises(ValueError, match="Both arrays should be of the same dimensions"):
```

jlpalomino: Good call. The test works without the `r`, so I have removed it now. Thanks, Max!

lwasser:
hey @jlpalomino is this PR complete / ready for a review? It seems to be but i didn't see a message that it was ready - just the slack ping.

jlpalomino:
@lwasser This PR is ready to be merged. Thanks!

lwasser:
woo hoo!! great work on this @jlpalomino !!! merging now.

PR #147, opened 2018-12-06
Tests for colorbar height and a few minor updates

lwasser:
* NOTE: still getting and odd mpl warning which i can't make sense of. Things still run however the use("agg") is throwing a warning consistently. might try to remove it and see what happens but first want to check this test with travis.


lwasser:
ok this includes 2 new colorbar tests. Also I renamed a few objects to make the entire file a bit more clear. I decided to create a new array for the colorbar tests as that will be used to plot only a single band (classified image).

mbjoseph:
Happy to review @lwasser - checking this out now!

lwasser:
Thank you for being awesome @mbjoseph !!! 😊

Leah

earthpy/tests/test_plot.py
```diff
@@ -16,7 +15,7 @@
 
 # Create tuple
```
mbjoseph: Should we delete this comment?

earthpy/tests/test_plot.py
```diff
@@ -16,7 +15,7 @@
...
 tuups = (1, 2)
-im = np.random.randint(10, size=(2, 4, 5))
+im_arr = np.random.randint(10, size=(2, 4, 5))
```
mbjoseph: Instead of defining this as a module-level variable, it would be better to use a pytest fixture, which ensures that each test function gets it's own version of the object (and in this case, different randomly generated numbers). You could do something like: 

```python
@pytest.fixture
def image_array_3d():
    return np.random.randint(10, size=(2, 4, 5))
```

earthpy/tests/test_plot.py
```diff
@@ -5,7 +5,6 @@
 import earthpy.spatial as es
 # For builds on travis to avoid plot display errors
 import matplotlib as mpl
```
mbjoseph: Looks like `mpl` is not called anymore - can we delete lines 6 and 7?

earthpy/tests/test_plot.py
```diff
@@ -16,7 +15,7 @@
 
 # Create tuple
 tuups = (1, 2)
```
mbjoseph: Instead of defining `tuups` as a module-level variable, it may be better to create such a tuple in the test that it is used (`test_arr_parameter` on line 20).

earthpy/tests/test_plot.py
```diff
@@ -16,7 +15,7 @@
...
-im = np.random.randint(10, size=(2, 4, 5))
+im_arr = np.random.randint(10, size=(2, 4, 5))
 
 def test_arr_parameter():
```
mbjoseph: Looks like this test can be simplified by not using the global variable `tuups`, e.g., 

```python
def test_arr_parameter():
    """Raise an AttributeError if an array is not provided."""
    with pytest.raises(AttributeError):
        es.plot_bands(arr=(1, 2))
```

earthpy/tests/test_plot.py
```diff
@@ -28,27 +27,27 @@ def test_num_titles():
     should raise an error OR if the title list is a different length than
     the array it should also raise an errors"""
 
-    single_band = im[0]
+    single_band = im_arr[0]
```
mbjoseph: Here `im_arr` is being pulled from the environment outside of the function implicitly, which can make it harder to understand the test. If we used a fixture for `im_arr` (e.g., the `image_array_3d` fixture suggested above), then the test could be rewritten in a way that makes it explicit what objects are required for the test to run: 

```python
def test_num_titles(image_array_3d):
    """If a user provides two titles for a single band array, the function
    should raise an error OR if the title list is a different length than
    the array it should also raise an errors"""
    single_band = image_array_3d[0]

        es.plot_bands(arr=single_band,
                      title=["Title1", "Title2"])
        es.plot_bands(arr=image_array_3d,
                      title=["Title1", "Title2", "Title3"])
```

earthpy/tests/test_plot.py
```diff
@@ -28,27 +27,27 @@ def test_num_titles():
...
 
     with pytest.raises(ValueError):
         es.plot_bands(arr=single_band,
                       title=["Title1", "Title2"])
     with pytest.raises(ValueError):
-        es.plot_bands(arr=im,
+        es.plot_bands(arr=im_arr,
                       title=["Title1", "Title2", "Title3"])
 
 def test_num_axes():
```
mbjoseph: Same comment here - the test uses a module-level data object that could be a fixture: 

```python
def test_num_axes(image_array_3d):
    """If provided with a 2 band array, plot_bands should return 3 axes by
    default"""
    fig, ax = es.plot_bands(image_array_3d)
    assert len(fig.axes) == 3
```

earthpy/tests/test_plot.py
```diff
@@ -28,27 +27,27 @@ def test_num_titles():
...
     """If provided with a 2 band array, plot_bands should return 3 axes by
     default"""
-    fig, ax = es.plot_bands(im)
+    fig, ax = es.plot_bands(im_arr)
     assert len(fig.axes) == 3
 
 
 def test_two_plot_title():
```
mbjoseph: Using a fixture here could improve clarity, and the `# Get titles` comment seems unnecessary, because of the `.get_title()` method, e.g.,: 

```python
def test_two_plot_title(image_array_3d):
    """Test that the default title is provided for a 2 band array plot"""
    fig, ax = es.plot_bands(image_array_3d)
    ax = fig.axes
    num_plots = image_array_3d.shape[0]
    plot_titles = [ax[i].get_title() for i in range(num_plots)]
    assert plot_titles == ['Band 1', 'Band 2']
```

earthpy/tests/test_plot.py
```diff
@@ -88,3 +87,29 @@ def test_single_band_2dims():
     assert arr.ndim == 2
     assert len(fig.axes[0].get_images()) == 1
 
+""" Colorbar Tests """
+
+
+im_arr2 = np.random.randint(10, size=(5, 5))
```
mbjoseph: It seems like `im_arr2`, `fig`, `ax`, and `im` could be created within the `test_colorbar_height` function instead, since they are only used there. 

Also (minor nits to pick):

- Is there a better name than `im_arr2`?
- Is specifying a `cmap` necessary for the test?

earthpy/tests/test_plot.py
```diff
@@ -88,3 +87,29 @@ def test_single_band_2dims():
...
+fig, ax = plt.subplots(figsize=(5, 5))
+im = ax.imshow(im_arr2,
+               cmap='RdYlGn')
+
+def test_colorbar_height():
```
mbjoseph: This test looks like it's written for a user-facing function with the try/except logic, but an `assert` statement would be more typical for a pytest test, e.g., 

```python
def test_colorbar_height():
    """Test that the colorbar ax height matches the image axis height"""
    fig, ax = plt.subplots(figsize=(5, 5))
    im = ax.imshow(np.random.randint(10, size=(5, 5)))
    cb = es.colorbar(im)
    assert cb.ax.get_position().height == im.axes.get_position().height
```

earthpy/tests/test_plot.py
```diff
@@ -88,3 +87,29 @@ def test_single_band_2dims():
...
+        cb.ax.get_position().height == \
+        im.axes.get_position().height
+    except AssertionError:
+        raise AssertionError("""Colorbar height does not match 
+                             the axes object provided.""")
+
+def test_colorbar_raises_value_error():
+    """Test that a non matbplotlib axis object raises an value error"""
+
+    with pytest.raises(AttributeError):
```
mbjoseph: This could use the `match` argument to make sure we get the right `AttributeError`, e.g., 

```suggestion
    with pytest.raises(AttributeError, match="requires a matplotlib"):
```

See https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions for more info

earthpy/tests/test_plot.py
lwasser: done!

earthpy/tests/test_plot.py
lwasser: ok i think i've fixed this throughout this test module.

earthpy/tests/test_plot.py
lwasser: done!

earthpy/tests/test_plot.py
lwasser: sure. i just created a tuple in place for this particular example. didn't seem to warrant an entire fixture given it's just a few characters. see what you think! :)

earthpy/tests/test_plot.py
lwasser: yes!! that is what i decided to do through. thank you for the suggestion!!

earthpy/tests/test_plot.py
lwasser: i'll be honest... i'm slightly afraid of putting fig objects in fixtures right now. i haven't done it so far. i want to be able to explicetly close them and know i can from a function but feel a bit bettter with them outside of the function. Particularly because i want to grab different parts of the object for diff tests. see what you think when i push these changes. i've also had problems trying to send TWO fixtures to a single test. in this case i'd want the array and the figure fixture. anyway see what you think when i push.  i'm open to whatever is the best approach.

earthpy/tests/test_plot.py
lwasser: sounds good! i do'nt think i need any comments in this module but wasn't sure if other people who are newer to MPL objects might find the comments helpful. i'd rather remove them all as I think the methods in the obj are pretty darn expressive!!! i removed this! 

as for the fixture for plot elements i'm going to push this and maybe we can chat briefly on slack.

earthpy/tests/test_plot.py
lwasser: i've created a fixture for the arr. I think it's a good idea to specify a cmap because it just is there for me to know colorbar works. all of the testing of creating the cbar should happen in mpl. But i think testing that the arguments that you think work still work is good even if there isn't an explicit test. i'd like to keep the cmap argument. i suppose i could create a new test for it but that is not a good idea because it opens a can of works of testing colorbar which again is a mpl function that is already (i hope) tested.

earthpy/tests/test_plot.py
lwasser: i have been writing too many front end functions apparently. i have no idea what i was thinking. fixed.

earthpy/tests/test_plot.py
mbjoseph: Word - I was suggesting that the array is passed as a fixture, rather than the figure. I'm not sure how passing figures as fixtures would work, but the changes you made to use fixtures for numpy arrays look great to me!

earthpy/tests/test_plot.py
mbjoseph: This suggestion was to use `image_array_3d` as a fixture, rather than using fixtures for plot elements. So, I think we are on the same page looking at the changes you had made!

earthpy/tests/test_plot.py
mbjoseph: Cool - thanks for clarifying!

mbjoseph:
Looks great to me @lwasser! I was suggesting that the arrays are passed as fixtures, but not that plot elements should be fixtures. So, I think we are on the same page!

lwasser:
ok wondeful. Thank you so much. I'll merge this @mbjoseph . I'm really open to plot fixtures in the future if we decide they are better.  yay! another PR merged today :)

PR #149, opened 2018-12-08
[WIP] rework of legend function p1

lwasser:
Lots of issues but want to begin to test on travis and make sure dif looks right

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -4,6 +4,7 @@ import numpy as np
 import numpy.ma as ma
 import matplotlib.pyplot as plt
 from matplotlib import patches as mpatches
+from matplotlib.colors import ListedColormap
 # For color bar resizing
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from shapely.geometry import mapping, box
@@ -595,7 +596,35 @@ def hillshade(arr, azimuth=30, angle_altitude=30):
     return 255*(shaded + 1)/2
 
 
-def draw_legend(im, classes, titles, bbox=(1.05, 1), loc=2):
+def make_col_list(unique_vals, cmap='Greys', nclasses=None):
+    """
+    Take a defined matplotlib colormap, and create a list of colors based on
+    a set of values. This is useful when you need to plot a series of
+    classified numpy arrays that are missing some of the sequential classes.
+    """
+    if not nclasses:
+        nclasses = len(unique_vals)
+
+    #TODO i think i can simplify this to
+    #cmap=plt.cm.get_cmap('Blues', 3)
+    cm = plt.cm.get_cmap(cmap)
+    increment = 1/(nclasses-1)
+
+    # Create increments to grab colormap colors
+    col_index = [(increment * c) for c in range(nclasses-1)]
+    col_index.append(1.0)
+    # Create cmap list
+    cols = [cm(c) for c in col_index]
+
+
+    return [cols[i-1] for i in unique_vals]
+
+
+def draw_legend(im_ax,
+                titles=None,
+                bbox=(1.05, 1),
+                cmap='Greys',
+                classes=None):
     """Create a custom legend with a box for each class in a raster using the
        image object, the unique classes in the image and titles for each class.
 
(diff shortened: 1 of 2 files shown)


lwasser:
This is also ready for review! @joemcglinchy  do you have any time to test using this function? i'd love some feedback. i have an example notebook if that would be helpful

lwasser:
thank you so much @joemcglinchy can you kindly redo your test case above creating an numpy array that i can recreate that is similar to your image / data? i'd be greatly appreciate and am greatly appreciate of your testing this function out!! so something like what is below that mimics the spread of the data in your image would be wonderful but still returns errors. Also i wonder 

```
class_bins = [-np.inf, 5, 10, 15, 18, np.inf]
classes=[1, 2, 3, 4, 5]

im_arr = np.random.randint(20, size=(15, 15))
ndvi = np.digitize(im_arr, class_bins)
```
if you know what i mean? then i can test against the data to understand what is going wrong!! alternatively you can send me a small piece of the image? but it would be nice to have something reproducible to work with. this is so helpful - thank you again.

lwasser:
just a question - for the example where ou provided titles, it should have thrown and error like the image below! it did for me. i would hope it would for you but please say the word if it's not. this is why i'd love to play with something more similar to your data. i am not yet telling you how the function works because it's helpful for me to see your response to it intuitively :) 

What i should probably do here is return a more useful error like. Your image as plotted only has 5 classes but you provided 10 titles. Consider specifying the classes that you expect in your image. AND be sure to normalize your data to the expected range of categories before plotting. 

screen shot 2018-12-11 at 11 45 40 am

joemcglinchy:
^^ in this case @lwasser the classes parameter was not provided. Thus it was calculated via np.unique() within the function and since the number of unique values calculated is the same as the number of labels there was no error.

joemcglinchy:
unfortunately I am unable to reproduce the missing legend with a random array implementation, but this code still produces the colormap differences:

```

class_bins = [-100, -0.8, -0.6, -0.4, -0.2, 0.0, 0.2, 0.4, 0.6, 0.8]
class_labels = ['nodata', 'neg 1', 'neg 2', 'neg 3', 'neg 4', 'neg 5', 'pos 1', 'pos 2', 'pos 3', 'pos 4']


im_arr = a = np.random.randn(15,15)

# the masked array assignments attempt to repro the ndvi example. feel free to remove
im_arr[im_arr < -1.] = -9999 # can remove
im_arr = np.ma.masked_equal(im_arr, -9999) # can remove
im_arr = np.clip(im_arr,-1,1)
im_arr_bin = np.digitize(im_arr, class_bins)
im_arr_bin = np.ma.masked_equal(im_arr_bin, 0) # can remove

# returns category X labels in legend, default imshow cmap (viridis) repeating colors
ax = plt.imshow(im_arr_bin)
es.draw_legend(ax)
plt.show()

# returns no visible legend in ndvi example, but works with random array
ax = plt.imshow(im_arr_bin )
es.draw_legend(ax,titles=class_labels)
plt.show()

# this plots with grayscale in legend but not matching CMAP in plt.imshow (default viridis)
ax = plt.imshow(im_arr_bin )
es.draw_legend(ax,classes=list(range(len(class_bins))), titles=class_labels)
plt.show()

# this plots with grayscale in legend but not matching CMAP in plt.imshow (default viridis)
ax = plt.imshow(im_arr_bin )
es.draw_legend(ax,classes=list(range(len(class_bins))))
plt.show()
```

lwasser:
ok! how big is that image @joemcglinchy ?? perhaps you can share it with me? i have to run my pup to the vet but will work on this more later. i found the bug that was causing the cmap color issue AND also things weren't being scaled properly. now i need to test
1. what happens when you provide this with a listed unnamed colormap? I think you will have to provide the colors to draw_legend.
2. why wasn't that legend plotting for you?!! 
thank you for testing this.

lwasser:
ok @joemcglinchy will you give this another test drive please? i've cleaned up a lot but want to see if it works better for you now!!

joemcglinchy:
@lwasser i ran my NDVI image through again. The colormap for the last 2 plots now matches the colormap provided in the call to imshow so that is great! I think the missing legend was due to me providing a number of classes that was a mismatch for the number of labels. The plot would still draw but the legend was not able to draw. I think this is good to go!

lwasser:
ok I need two things on this PR
@joemcglinchy will you kindly test out the legend code again now? Please note that for the time being this function does NOT support side by side unique legends. I know this is something some people might want but it wasn't in the original scope of the function. So... i suggest for the time being we test that it works

1. for all types of masked and non masked single axis plots
2. it should actually work for two plots that are stacked on top of each other.

Please have a look and let me know if it works now consistently across those above use cases. Then we can create a new issue to add the side by side functionality which will get a bit tricky as i think i'll have to create a new divider element or something. not sure. 

@mbjoseph if you have time to review this PR for testing and consistency it would be wonderful as well.

lwasser:
@mbjoseph also this PR has the examples of where i wanted to use two fixtures in a single test. You can see that the tests fail. I'd love your guidance on this. you can see where i commented out the second fixture import as it doesn't seem to work this way.

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -0,0 +1,223 @@
...
+@pytest.fixture
+def binned_array():
+    im_arr = np.random.uniform(-2, 1, (15, 15))
+    bins = [-100, -0.8, -0.2, 0.2, 0.8, np.Inf]
+    im_arr_bin = np.digitize(im_arr, bins)
+    return bins, im_arr_bin
+
+
+@pytest.fixture
+def listed_cmap():
```
mbjoseph: There is already a fixture called `listed_cmap()` defined above this - maybe a whoopsie? The first one returns a tuple, and this returns just one thing. It looks like that's whats causing problems below @lwasser - maybe if we stick with the first `listed_cmap` instead of redefining it here, the tuples below should be unpacked as expected?

mbjoseph:
Looks good to me @lwasser!

Issue #142, opened 2018-12-06
Executing example code in Read the Docs documentation


lwasser:
Can we setup doc builds so that example code within the .rst files is executed?
[Example doc with code](https://github.com/earthlab/earthpy/blob/master/docs/spatial-raster.rst#calculate-normalized-difference) 
Using doctest

mbjoseph:
Solved by #151

PR #151, opened 2018-12-11
Run example code in docs with doctest

mbjoseph:
This PR adds the doctest Sphinx module, and uses Travis CI to execute example code in rst files and docstrings as proposed in #142. 

It also adds some small datasets (1.5 MB total) that can be used in examples, including:

- RGB satellite imagery over Rocky Mountain National Park (RMNP) and vicinity as a 3 band GeoTIFF, and also an "unstacked" version - one GeoTIFF file per band
- Digital elevation model of RMNP and surrounding area
- Polygon boundary of RMNP
- Path of the Continental Divide Trail through the U.S. (including through RMNP)
- Point locations of glaciers in Colorado (including some in RMNP)

## Examples to fix

- [x] docs/spatial-raster.rst: Examples here use non-existent GeoTIFF files and directories, and we also need to include expected output for the examples to verify that the output is what we expect. 

- [x] docs/spatial-vector.rst: These examples use non-existent shapefiles, and also need expected output to make sure the examples work. 

- [x] earthpy/mask.py: This example seems to have a broken import section, and needs expected output. 

- [x] earthpy/spatial.py: These examples need expected output.

2018-12-17: Edited to describe example datasets


lwasser:
hey @mbjoseph this is great! i think it would be nice to include a small data subset like what rasterio uses. i mean we could maybe even use their data? they have a nice small dataset that i presume is what they use in their builds. then the mask function for instance might make a bit more sense - if it's run on images.

we have data already in earthpy
https://github.com/earthlab/earthpy/tree/master/earthpy/data 
The data are accessed here: https://github.com/earthlab/earthpy/blob/master/earthpy/__init__.py#L15 but i'm not sure exactly how that differs from calling data (or if it does) 

The includes could be a nice way to utilize the data in the docs !! i just haven't ever done this  before to know how it works or provide guidance. 

i am happy to adjust the docs however to support this pr. 
Would you like to begin with some examples that are completely reproducible? So i could make and plot a GDF for example and add that to the docs. that would run without data imports. 

just let me know what would be best for me to do to help build out this PR!

lwasser:
ps. i noticed last night that our docs are failing according to the badge. it's weird since travis seems happy.

lwasser:
UPDATE: just fixed the build. not sure what happened there. Anyway lemme know what you need here!! Happy to help!

mbjoseph:
Cool - thanks @lwasser for the examples from rasterio. It looks like those need to be executed from a specific working directory, e.g., 

```python
import rasteriopysal examples
src = rasterio.open("tests/data/RGB.byte.tif")
```

This seems fine for testing whether the example code will run and is dev-friendly in terms of setting up doctest, but is not very user friendly because it assumes the code is being run from a specific directory (we may have to explain to a student that they need to clone the repo and execute the code using a working directory in the source code of earthpy!). 

Pysal has a more portable solution for packaging example data - they have a `pysal.examples` module that allows users to access example data like this, regardless of the current working directory (taken from https://pysal.readthedocs.io/en/latest/users/tutorials/fileio.html): 

```python
import pysal
shp = pysal.open(pysal.examples.get_path('10740.shp'))
```

Each dataset is a subdirectory of `pysal/examples` in their repo: https://github.com/pysal/pysal/tree/master/pysal/examples

And they have utility functions for access here: https://github.com/pysal/pysal/blob/master/pysal/examples/__init__.py

I will work on something similar with our data directory so that we (and users) can call example data stored in earthpy with this kind of syntax:

```python
import rasterio rio
import earthpy.data as ed
with rio.open(ed.get_path('RGB.byte.tif')) as src:
    # cool example here
```

lwasser:
ahhh that's cool. Our get_data method .. i wonder if any of the code in that io module could be useful here because essentially it's a similar idea except it finds a home dir (default option) on the users computer. different application but similar concept. 

Oddly the data element - i swear we used the json file identified in the `__init__` as a lookup somewhere but it is gone now. The goal was to lookup a prj string from the epsg code. Now, I can't seem to find it being called / used anywhere in our package.  i'll make a new issue for that.

lwasser:
ok now i remember. the way it's setup is you can call

et.epsg["epsg-code-here"] so any user can access the json file and use it to look up a prj string. it's totally undocumented however!

earthpy/clip.py
```diff
@@ -119,7 +119,7 @@ def clip_shp(shp, clip_obj):
                              geometry column""")
 
 
-    if not any(shp.intersects(clip_obj)):
+    if not any(shp.intersects(clip_obj.unary_union)):
```
mbjoseph: This was a fix that was necessary for the function to work when clipping spatial points by a spatial polygon (the example in `docs/spatial-vector.rst` where Colorado glacier point locations are clipped by the boundary of Rocky Mountain National Park).

earthpy/data/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+A dictionary that maps numeric EPSG codes to their proj4string representations.
+
+### Filename
+
+`epsg.json`
+
+### Source
+
+TODO: name source
```
mbjoseph: Not sure where this data was acquired, but we can put the source here

earthpy/data/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+### License
+
+TODO: include license info
```
mbjoseph: Did this dataset have any licensing restrictions that we should list here?

earthpy/spatial.py
```diff
@@ -266,10 +274,10 @@ def bytescale(data, cmin=None, cmax=None, high=255, low=0):
     if high < low:
         raise ValueError("`high` should be greater than or equal to `low`.")
 
-    if (cmin is None) or (cmin < data.min()):
+    if cmin is None:
```
mbjoseph: This line reverts back to the original scipy implementation. The change for these conditional statements was necessary for the scipy examples to provide the expected results, and did not break any of our existing unit tests. I looked at the pull request (https://github.com/earthlab/earthpy/pull/145) where we added `or (cmin < data.min())` and `or (cmax > data.max())` and couldn't tell why the `or` statements were added. 

It would be great if @joemcglinchy could take a look at this change and make sure this change doesn't break something (and if it does, we should add a unit test to catch that edge case). But, the existing earthpy bytescale function did not generate the expected output for the third example in the bytescale docstring which was taken from the original scipy docstring - it was scaling values to have range (0, 255) whereas the expected output from the original scipy version has range (3, 91). 

My assumption here is that we want this function to behave like scipy's bytescale function, but if that's wrong, let me know!

earthpy/spatial.py
```diff
@@ -463,7 +473,7 @@ def plot_rgb(arr, rgb=(0, 1, 2),
                            order with bands first""")
 
     # Index bands for plotting and clean up data for matplotlib
-    rgb_bands = arr[rgb]
+    rgb_bands = arr[rgb, :, :]
```
mbjoseph: This change I think fixes a bug in `plot_rgb`. Previously this line would return one value for the default value of the `rgb` argument, because a 3 dimension ndarray was being indexed using a length 3 tuple, e.g.,:

```python
>>> import numpy as np
>>> arr = np.arange(9).reshape((1, 3, 3))
>>> rgb = (0, 1, 2)  # the default value for the rgb parameter
>>> rgb_bands = arr[rgb] 
>>> rgb_bands
5
```

@lwasser please take a close look at this change and make sure I'm understanding correctly. I'm not sure whether this function is in active usage in your class, so maybe I've missed something and this is not actually a bug.

earthpy/spatial.py
```diff
@@ -405,7 +417,6 @@ def plot_bands(arr, title=None, cmap="Greys_r",
             ax.set_axis_off()
             ax.set(xticks=[], yticks=[])
         plt.tight_layout()
-        plt.show()
```
mbjoseph: This call to `plt.show()` seems fairly restrictive. Is it totally necessary? It seems like removing it would allow people to create one plot with multiple arrays, e.g., 

```python
>>> import numpy as np
>>> import earthpy.spatial as es
>>> im = np.random.randint(10, size=(2, 4, 5))
>>> titles = ["Red Band", "Green Band"]
>>> es.plot_bands(im,
...               title=titles,
...               figsize=(12,5),
...               cols=2)
>>> es.plot_bands(im,
...               title=titles,
...               figsize=(12,5),
...               cols=2) 
plt.show()
```

Also, by removing these calls to `plot_bands`, we won't need to worry about interactively closing a bunch of plotting windows when running tests locally. So we get more flexibility for users, and less clicking around for the devs!

earthpy/spatial.py
joemcglinchy: based on their documentation i would expect the third example to have output with range (0,255) the same as the first example, as opposed to directly converting the input array to uint8 by way of rounding. What their example seems to is say if the data is already in the range (0,255), ie is of type uint8, then do no scaling regardless of how the cmin/cmax parameters are specified. Currently, we are saying that if the cmin or cmax is less than or greater than the data.min() or data.max(), then use the data.min or data.max(). I think that is correct, but the only difference is we are still scaling the data in that case to the values specified by `low` and `high`. In other words, we are always stretching the data to be between 0-255 even if the data is already in that range. We could add a conditional that says if the data is already in that range, then just return the data with an `.astype('uint8')`, which would be appropriate.

mbjoseph: Cool - thanks @joemcglinchy! I agree that based on their documentation I would expect their third example to generate a range of values from (0, 255), rather than the original data range. 

So, we need to decide whether to stick with the original scipy behavior, which is: 

```python
>>> img = np.array([[ 91.06794177,   3.39058326,  84.4221549 ],
...                 [ 73.88003259,  80.91433048,   4.88878881],
...                 [ 51.53875334,  34.45808177,  27.5873488 ]])

>>> bytescale(img, cmin=0, cmax=255)
array([[91,  3, 84],
       [74, 81,  5],
       [52, 34, 28]], dtype=uint8)
```

or whether we want to rescale the values to have range (0, 255), even when the original data range is already contained within that range (the behavior after #145, and incidentally also the behavior of `gdal_translate`):

```python
>>> img = np.array([[ 91.06794177,   3.39058326,  84.4221549 ],
...                 [ 73.88003259,  80.91433048,   4.88878881],
...                 [ 51.53875334,  34.45808177,  27.5873488 ]])

>>> bytescale(img, cmin=0, cmax=255)
array([[255,  0, 237],
       [206, 226,  4],
       [141, 91, 71]], dtype=uint8)
```

I would advocate for the behavior in the second example, rather than the original scipy behavior, because it is more consistent with the documentation, and is identical to `gdal_translate` byte scaling. Does that sound reasonable @lwasser? If so, I will change the example in the docstring.

earthpy/__init__.py
```diff
@@ -6,7 +6,7 @@
 import os.path as op
 from . import utils, spatial
 
-data = EarthlabData()
+earthlabdata = EarthlabData()
```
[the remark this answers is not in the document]
lwasser: @mbjoseph i like .data as it keeps the call simple. May i ask why you renamed it earthlab data? the idea with this is it doesn't have to be earthlab data. it can be any data. we've updated the class to accept url's now too to facilitate using the hub. so i'd rather not make this change.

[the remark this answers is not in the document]
lwasser: Here is a page where we call to get the data
https://www.earthdatascience.org/courses/earth-analytics-python/multispectral-remote-sensing-in-python/ but we use it on every lesson everywhere!!

you can also call this:

`et.data.get_data(url="https://ndownloader.figshare.com/files/12459464")` which again allows you to get ANY data and unpack it in the same way that you can unpack our prepackaged data. My goal for this function is it's a way for a class  using earthpy to make it easier for students to get data into an established working dir. 

We could consider renaming the `earthpy/data` directory rather than the instance defined in the init file in this case.

earthpy/clip.py
lwasser: oh gosh good catch. yes you'd need unary_union to ensure that the full extents overlap. thank you!!

earthpy/data/README.md
lwasser: Should this go in our docs too? i suppose we can add it as an include especially if we use rst.

lwasser: EPSG data came from here via Chris H. :
```
# This EPSG mapping converted from:
# https://github.com/jswhit/pyproj/blob/master/lib/pyproj/data/epsg
```

earthpy/data/README.md
lwasser: i don't know!! none of it is our data explicitly.

earthpy/spatial.py
lwasser: i also advocate for parallelling `gdal_translate`. it makes the most sense to stretch the values between 0 and 255 as that is what a user would expect when plotting!! Great catch and thank you so much!!

earthpy/spatial.py
lwasser: hmmm definitely confused here. this function did work just fine before but i think i need to dig into it to understand what the bug is you are mentioned. i do remember seeing a deprecation of subsetting the way we were doing it...

earthpy/__init__.py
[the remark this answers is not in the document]
mbjoseph: Cool thank you @lwasser! I suspected it was being used already. I agree renaming the current `earthpy/data/` directory seems like a good solution (and a way to avoid confusion for others, who would expect `earthpy.data` to relate to the `earthpy/data/` directory).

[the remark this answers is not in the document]
mbjoseph: Reverted back to `earthlab.data` in b85c98482b504f3704d3693fc72bafc774b7c952

earthpy/clip.py
mbjoseph: :+1:

earthpy/data/README.md
mbjoseph: Thanks! Added this in 944b8b37f88ee8b0abc3719ab1efbb37bc05be9c

earthpy/data/README.md
mbjoseph: Thanks! I inserted the relevant info in 944b8b37f88ee8b0abc3719ab1efbb37bc05be9c

earthpy/spatial.py
mbjoseph: Cool - thanks again @joemcglinchy and @lwasser! Commit a11479a956403f7d1e8e9c3dfd1a3979e092edb9 makes the bytescale function act like we want.

earthpy/spatial.py
mbjoseph: Solved IRL!

earthpy/spatial.py
mbjoseph: Fixed this IRL!

mbjoseph:
Ok @lwasser this is ready to review! When we merge, let's try remember to squash the commits from this branch if possible.

mbjoseph:
Heya @lwasser merge conflicts are fixed on this PR, so should be ready to review!

lwasser:
wow. this is an awesome pr @mbjoseph thank you so much! merging this (and squashing) now!!

Issue #155, opened 2018-12-13
stack_raster_tifs cannot write tifs to the current directory


mbjoseph:
The `stack_raster_tif` function raises an error when writing output to the current working directory, e.g., 

```python
arr, arr_meta = es.stack_raster_tifs(paths, "out.tif")
```

This error occurs because of this line (https://github.com/earthlab/earthpy/blob/master/earthpy/spatial.py#L115):

```python
if not os.path.exists(os.path.dirname(out_path)):
    # Raise a value error
```

If `out_path` is "out.tif", then `os.path.dirname(out_path)` evalues to `""`, and `os.path.exists("")` evaluates to `False`. 

I have a fix in the works as part of #151, because this issue came up in the example code.

mbjoseph:
Solved by #156

PR #156, opened 2018-12-13
Allow writing stack output to current working directory

mbjoseph:
This PR allows users to write output from `stack_raster_tifs` to their current working directory, e.g., 

```python
import earthpy.spatial as es
es.stack_raster_tifs(band_paths=['foo.tif', 'bar.tif'], out_path='foobar.tif')
```

Previously this would raise an error as described in #155. @jlpalomino, @joemcglinchy, or @lwasser is somebody willing to review?


lwasser:
on it @mbjoseph  We all have the same PR conflicts now. We'll have to clean that up this week.

earthpy/spatial.py
```diff
@@ -112,7 +112,9 @@ def stack_raster_tifs(band_paths, out_path, arr_out=True):
     # Set default import to read
     kwds = {'mode': 'r'}
 
-    if not os.path.exists(os.path.dirname(out_path)):
+    out_dir = os.path.dirname(out_path)
+    writing_to_cwd = out_dir == ""
```
lwasser: @mbjoseph i have a hard time understanding this line. doesn't this set both vars to "" ? I know there is just something i dont understand here.

the `out_dir` parameter was previous `out_path` in hopes users would understand they need to provide a path to the directory not just the name of it. This was confusing for students

earthpy/tests/test_spatial.py
```diff
@@ -114,37 +114,44 @@ def test_extent_to_json():
...
+
     # Test scale value max is less equal to min
     scale_arr = es.bytescale(arr, cmin=10, cmax=240)
-    
+
     assert scale_arr.min() == 0
     assert scale_arr.max() == 255
+
+
+def test_stack_invalid_out_paths_raise_errors():
```
lwasser: @mbjoseph do we also need to add a test for a user trying to write to the CWD? essentially what you are fixing here in this PR?

earthpy/spatial.py
mbjoseph: Hi @lwasser - not quite: `out_dir` will be a string, and `writing_to_cwd` will be boolean because of the `==` operator.

```python
>>> out_dir = os.path.dirname("output.tif")
>>> out_dir
''
>>> writing_to_cwd = out_dir == ""
>>> writing_to_cwd
True
```

earthpy/tests/test_spatial.py
mbjoseph: There are a lot of tests we need to write for stack_raster_tifs - this test is just a start!

lwasser: ok wonderful. @joemcglinchy is working on these tests. my apologies i am forgetting the intent of this PR!

earthpy/spatial.py
lwasser: Oh. i was reading this wrong. there is a double `==`. my brain has a hard time reading all of the equals in that line. i get it now. thank you.

earthpy/tests/test_spatial.py
joemcglinchy: excuse me if I have missed this. Does the update allow for writing to CWD??

mbjoseph: @joemcglinchy this PR allows users to write to the CWD by specifying an output path as `"output.tif"`, instead of `"./output.tif"`. The latter is the only way to do it currently, and both will be possible after merging this PR.

mbjoseph:
Ok @lwasser merge conflicts are fixed - I think this one may be ready to merge!

earthpy/tests/test_spatial.py
lwasser: 💯

lwasser:
yahoo!! thank you again @mbjoseph !!

PR #158, opened 2018-12-16
Add black as a pre-commit hook

### LINKED ISSUE #132: Automate linting
It would be nice to automatically lint in our continuous integration pipeline. We can discuss what to use here, but some options include [pycodestyle](https://github.com/PyCQA/pycodestyle), [pylint](https://github.com/PyCQA/pylint), and [flake8](https://github.com/PyCQA/flake8). We could also look into something that automatically formats code like [black](https://github.com/ambv/black).

Let's use this issue to decide what we want to use.

betatim:
This adds `black` as a pre-commit hook.

I followed the link to the blog post Joe linked to and then ended up on https://github.com/ambv/black#version-control-integration which uses the same tool (`pre-commit`) to set things up. Conclusion: I just followed the instructions 😄.

I didn't add `flake8` as a pre-commit hook as I think `black` will reformat changes you made that would violate pep8 in black's steamroller way of dealing with things.

Closes #132

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -20,6 +20,10 @@ When submitting a change to the repository, please do the following:
 
 When submitting a pull request:
 
+- Install the tools needed for development by running
+  `pip install -r dev-requirements.txt` or the equivalent of it for your package
+  manager.
+
 - All existing tests should pass. Please make sure that the test
   suite passes, both locally and on
   `Travis CI <https://travis-ci.org/earthlab/earthpy>`_
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.black]
+line-length = 79
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+- repo: https://github.com/ambv/black
+  rev: stable
+  hooks:
+  - id: black
+    language_version: python3.6
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,4 @@
 sphinx==1.8.0
 sphinx-autobuild==0.7.1
 sphinx_rtd_theme
+pre-commit


betatim:
One thing that is a bit weird with this setup is that if you edit a single line in a file then black will reformat the whole file. This means your single line edit becomes huge. Checking if it is possible to configure black so that it only applies its reformatting to the lines you edited (or nearby).

That way you can slowly convert a file to the new style as you anyway edit stuff. Or maybe we make one PR that applies black to everything. I like keeping commits that do nothing but reformat separate from commits that change contents :-/

lwasser:
hey @betatim i do agree that it would be harder to have to review a code update that is a few lines with more lines of edits given formatting fixes via black. i'm definitely liking black more and getting used to seeing code formatted that way. It also means i don't have to ping people about syntax which is great :) and no fighting with travis. So if there is a way to keep format edits separate that would be ideal.

betatim:
Did a bit of searching around but couldn't find anyone wanting to only format "changed lines". Maybe it is hard to do? For example if I add a new argument to a function then just formatting that one line "according to black" would create pretty weird looking code if the rest of the arguments were still formatted the old way :)

Right now I am thinking: if there is consensus to go for black I'd make one PR that does nothing but run `black` on the whole project and merge that. After that has gone in having a commit hook like the one proposed here wouldn't create extra noise.

CONTRIBUTING.rst
```diff
@@ -20,6 +20,10 @@ When submitting a change to the repository, please do the following:
 
 When submitting a pull request:
 
+- Install the tools needed for development by running
```
mbjoseph: :+1:

lwasser:
I believe that we have a consensus here. is that right @mbjoseph  i just see your thumbs up :) if so i think we can merge this!! i'm using black more and more. it's just very easy to manage.

mbjoseph:
I think adding a pre-commit hook for black is a great idea - if I am reading @betatim's suggestion correctly:

> Right now I am thinking: if there is consensus to go for black I'd make one PR that does nothing but run black on the whole project and merge that. After that has gone in having a commit hook like the one proposed here wouldn't create extra noise. 

...then we want to wait to merge this until we merge a separate PR that incorporates all of the formatting changes that black will make in the existing code. But once we merge in the formatting changes, I think we are good to merge this!

lwasser:
oooh my apologies. you are totally right @mbjoseph  i misread this. I think that is a great idea. i've been running black on every file as i work in pycharm but you can do an entire directory easily too at the CL. good call.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 79
```
betatim: Before you hit merge: the `black` default is 88. Which is probably what got used in #161 

I think if you aren't as stuck in your ways as I am 88 is a good choice (and simplifies things).

lwasser: oh... you know i had thought i reran it using 79 but you are likely right @betatim 
@mbjoseph do you have thoughts on 79 vs 88 character line width? pycharm defaults to 88 with black. i'm still trying to figure out how to fix that. should we stick to pep8 convention? black actually does break a few things like the single vs double quote rules already. thoughts??

lwasser:
@mbjoseph that is a good question. i just looked at that file and it seems to only have reqs for docs. But i would think we want to have black as a requirement somewhere i'm just not sure where it should be. maybe there what is protocol here?

pyproject.toml
lwasser: ok i did fix this to be 79 last night @betatim @mbjoseph  -- in my last commit [here](564a8a83895ee53ac60c93cd402954bb416e9758)

we are in agreement that we'd like to use 79 for this repo (just chatted with max) SO... can we merge this?! i'm excited to have everyone apply this on their pr's and such too.

lwasser:
so stealing this from @mbjoseph

betatim:
> do we need to add black in the `dev-requirements.txt` file?

I think the precommit hook package does some magic to take care of installing black for the purposes of the pre-commit hook (in a different temporary environment??). It "just worked" for me.

mbjoseph:
Weird - I'm testing this branch locally on a fresh environment installed from earthpy's environment.yml file and I can commit all kinds of non-black things (single quotes, super long lines), and I'm not seeing any pre-commit hooks coming into play.

I pulled this branch and created a fresh environment via

```bash
conda env create -f environment.yml
conda activate earthpy-dev
pip install -r dev-requirements.txt
````

Do we need to do something like `pre-commit install` to activate the hook? Or, is there something else that we need to do to turn this on? Apologies if I missed something obvious.

lwasser:
super duper weird. this all works fine on my computer @mbjoseph  . (MAC OS)

PR #161, opened 2018-12-18
blackifying all the things

lwasser:
me likey black weeeee!!!


earthpy/clip.py
```diff
@@ -77,7 +77,7 @@ def clip_line_poly(shp, clip_obj):
 
     # Clip the data - with these data
     clipped = shp_sub.copy()
-    clipped['geometry'] = shp_sub.intersection(poly)
+    clipped["geometry"] = shp_sub.intersection(poly)
```
lwasser: @betatim should't dataframe cols be single quotes? thought that was the rule! maybe i am wrong.

earthpy/mask.py
```diff
@@ -22,18 +22,66 @@
...
+                432,
+                480,
+            ],
             "Medium Cirrus Confidence": [],
-            "High Cirrus Confidence": [834, 836, 840, 848, 864, 880, 898, 900,
-                                       904, 912, 928, 944, 992],
-            "Terrain Occlusion": [1346, 1348, 1350, 1352]
-        }
+            "High Cirrus Confidence": [
+                834,
```
lwasser: this makes me a bit sad given all of the lines but...

earthpy/spatial.py
```diff
@@ -110,34 +113,37 @@ def stack_raster_tifs(band_paths, out_path, arr_out=True):
...
+            with rio.open(out_path, "w", **dest_kwargs) as dest:
                 stack(sources, dest)
             # Read and return array
-            with rio.open(out_path, 'r') as src:
+            with rio.open(out_path, "r") as src:
                 return src.read(), src.profile
         else:
             # Write stacked gtif file
-            with rio.open(out_path, 'w', **dest_kwargs) as dest:
+            with rio.open(out_path, "w", **dest_kwargs) as dest:
```
lwasser: similarly @betatim i thought that arguments should be single quotes too.

earthpy/clip.py
lwasser: ok i get it. they like double quotes every.where. ok fine. :)

mbjoseph:
Heya @lwasser this is looking pretty good! One quick question: I noticed that black wasn't run on `./setup.py` or `docs/conf.py`: is that on purpose? Looking at Tim's PR, I expect that we will be running black from the top-level directory, which means it should hit these two files. 

If you execute `black .` from the top level earthpy directory, does it hit `setup.py` and `docs/conf.py` for you locally? If so, it seems like we could commit those changes to this branch too. After including changes to those files, I think we can merge!!

lwasser:
okdokey! @mbjoseph fixed PLUS i fixed the line length. changed to 79 to match Tim's pr!

lwasser:
how did i miss the SEAL of approval? 😆 !!!

PR #162, opened 2018-12-18
added try-except block to check for dataset reader

joemcglinchy:


lwasser:
thank you @joemcglinchy i will merge this now. it's failing code cov but we are workig on that independently. this will ensure that those with older versions of rasterio won't get an error message due to the change on reader object name!

PR #164, opened 2018-12-19
setup pypi updates

lwasser:
This simply updates the setup and readme to be a bit nicer on pypi.We will want to discuss how we document contributors vs maintainers, etc i think somewhere??


lwasser:
merging this as it's for pypi!!

PR #170, opened 2018-12-19
Stack raster tests jm

joemcglinchy:
i removed the original `stack_raster_tifs` -> `stack` workflow and replaced it with `stack` -> `stack_bands`. the old workflow functions have a warning saying what to use and exit immediately.

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -153,7 +153,7 @@ def stack(band_paths, out_path='', write_raster=False):
         # Stack the bands and return an array, but don't write to disk
         if not write_raster: 
             
-            arr, prof = stack_bands(sources)
+            arr, prof = _stack_bands(sources)
             return arr, prof
         
         # Write out the stacked array and return a numpy array    
@@ -168,14 +168,14 @@ def stack(band_paths, out_path='', write_raster=False):
             
             # Write stacked gtif file
             with rio.open(out_path, "w", **dest_kwargs) as dest:
-                stack_bands(sources, dest, write_raster)
+                _stack_bands(sources, dest, write_raster)
             
             # Read and return array
             with rio.open(out_path, "r") as src:
                 return src.read(), src.profile
         
                 
-def stack_bands(sources, dest=None, write_raster=False):
+def _stack_bands(sources, dest=None, write_raster=False):
     """Stack a set of bands into a single file.
 
     Parameters
@@ -220,6 +220,119 @@ def stack_bands(sources, dest=None, write_raster=False):
         
         return np.array(stacked_arr), ret_prof
 
+# @deprecate
+def stack_raster_tifs(band_paths, out_path, arr_out=True):
+    """Take a list of raster paths and turn into an output raster stack
+    numpy array. Note that this function depends upon the stack() function.
+
+    Parameters
+    ----------
+    band_paths : list of file paths
+        A list with paths to the bands you wish to stack. Bands
+        will be stacked in the order given in this list.
+    out_path : string
+        A path with a file name for the output stacked raster
+         tif file.
+    arr_out : boolean
+        A boolean argument to designate what is returned in the stacked
+        raster tif output.
+
+    Returns
+    ----------
+    If arr_out keyword is True:
+        tuple: The first value representing the result of src.read() of the
+        stacked array and the second value
+        representing the result of src.profile of the stacked array.
+    If arr_out keyword is False:
+        str : A path with a file name for the output stacked raster tif file.
+
+    TODO: Instead of returning a file path when arr_out=False, consider
+        returning None since the out_path is already
+        an input given by the user. This will make the output type consistent.
+    """
+    
+    # Throw warning and exit
+    raise Warning("stack_raster_tifs is deprecated. Use stack(). Exiting...")
+    sys.exit()
+    
+    
+    # Set default import to read
+    kwds = {"mode": "r"}
+
+    out_dir = os.path.dirname(out_path)
+    writing_to_cwd = out_dir == ""
+    if not os.path.exists(out_dir) and not writing_to_cwd:
+        raise ValueError(
+            "The output directory path that you provided does not exist"
+        )
+
+    if len(band_paths) < 2:
+        raise ValueError(
+            """The list of file paths is empty. You need at least
+                            2 files to create a stack."""
+        )
+    with contextlib.ExitStack() as context:
+        sources = [
+            context.enter_context(rio.open(path, **kwds))
+            for path in band_paths
+        ]
+
+        # TODO: Check that the CRS and TRANSFORM are the same
+        dest_kwargs = sources[0].meta
+        dest_count = sum(src.count for src in sources)
+        dest_kwargs["count"] = dest_count
+
+        if arr_out:
+            # Write stacked gtif file
+            with rio.open(out_path, "w", **dest_kwargs) as dest:
+                _stack(sources, dest)
+            # Read and return array
+            with rio.open(out_path, "r") as src:
+                return src.read(), src.profile
+        else:
+            # Write stacked gtif file
+            with rio.open(out_path, "w", **dest_kwargs) as dest:
+                return _stack(sources, dest)
+
+
+# Function to be submitted to rasterio
+# TODO: add unit tests - some are here:
+# https://github.com/mapbox/rasterio/blob/master/rasterio/mask.py
+# This function doesn't stand alone because it writes to an open object called
+# in the other function
+# @deprecate
+def _stack(sources, dest):
+    """Stack a set of bands into a single file.
+
+    Parameters
+    ----------
+    sources : list of rasterio dataset objects
+        A list with paths to the bands you wish to stack. Objects
+        will be stacked in the order provided in this list.
+    dest : a rio.open writable object that will store raster data.
+    """
+
+    # Throw warning and exit
+    raise Warning("_stack is deprecated. Exiting...")
+    sys.exit()
+    
+    try:
+        for src in sources:
+            src.profile
+
+    except ValueError as ve:
+        raise ValueError("The sources object should be Dataset Reader")
+        sys.exit()
+
+    else:
+        pass
+
+    for ii, ifile in enumerate(sources):
+        bands = sources[ii].read()
+        if bands.ndim != 3:
+            bands = bands[np.newaxis, ...]
+        for band in bands:
+            dest.write(band, ii + 1)
 
 
 def crop_image(raster, geoms, all_touched=True):
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -89,37 +89,31 @@ def normalized_diff(b1, b2):
 
 
 # TODO: include a no data value here if provided
+def stack(band_paths, out_path='', write_raster=False):
 
-
-def stack_raster_tifs(band_paths, out_path, arr_out=True):
     """Take a list of raster paths and turn into an output raster stack
-    numpy array. Note that this function depends upon the stack() function.
+    numpy array. Note that this function depends upon the stack_bands() function.
 
     Parameters
     ----------
     band_paths : list of file paths
         A list with paths to the bands you wish to stack. Bands
         will be stacked in the order given in this list.
-    out_path : string
+    out_path : string (optional)
         A path with a file name for the output stacked raster
          tif file.
-    arr_out : boolean
+    write_raster : boolean (optional)
         A boolean argument to designate what is returned in the stacked
         raster tif output.
 
     Returns
     ----------
-    If arr_out keyword is True:
-        tuple: The first value representing the result of src.read() of the
-        stacked array and the second value
-        representing the result of src.profile of the stacked array.
-    If arr_out keyword is False:
-        str : A path with a file name for the output stacked raster tif file.
-
-    TODO: Instead of returning a file path when arr_out=False, consider
-        returning None since the out_path is already
-        an input given by the user. This will make the output type consistent.
-    """
+    tuple: The first value representing the numpy array resulting from stacking the files in the input list
+        and the second value representing the result of src.profile of the stacked array.
+        NOTE: the 'count' key of the profile is updated to match the length of the input list.
+    If write_raster keyword is True:
+        a file will be written from the stacked array to the path specified in out_path."""
+        
     # Set default import to read
     kwds = {"mode": "r"}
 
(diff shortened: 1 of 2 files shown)


earthpy/spatial.py
```diff
@@ -168,6 +298,12 @@ def stack_raster_tifs(band_paths, out_path, arr_out=True):
...
         >>> os.remove(destfile)
     """
+    
+    # Throw warning and exit
+    raise Warning("stack_raster_tifs is deprecated. Use stack(). Exiting...")
+    sys.exit()
+    
+    
     # Set default import to read
     kwds = {"mode": "r"}
```
lwasser: Do we need to retain the code below the warning above given we are deprecating this function @joemcglinchy

earthpy/spatial.py
```diff
@@ -121,8 +121,138 @@ def normalized_diff(b1, b2):
...
+                bands = np.squeeze(bands)
+            stacked_arr.append(bands)
+        
+        # update the profile to have count==number of bands
+        ret_prof = sources[0].profile.copy()
+        ret_prof['count'] = len(stacked_arr)
+        
+        return np.array(stacked_arr), ret_prof
+
+# @deprecate
```
lwasser: Why don't we move this deprecated function to the bottom of this file. @mbjoseph do you have any thoughts on the location of deprecated functions within our package??? i was just thinking moving it down but am open to suggestions

earthpy/spatial.py
```diff
@@ -121,8 +121,138 @@ def normalized_diff(b1, b2):
...
+                
+    else:
+        stacked_arr = []
+        for ii, ifile in enumerate(sources):
+            bands = sources[ii].read()
+            if bands.shape[0] == 1:
+                bands = np.squeeze(bands)
+            stacked_arr.append(bands)
+        
+        # update the profile to have count==number of bands
```
lwasser: Please capitalize the first letter of the comment.

earthpy/spatial.py
```diff
@@ -197,22 +333,23 @@ def stack_raster_tifs(band_paths, out_path, arr_out=True):
...
 
 
 # Function to be submitted to rasterio
 # TODO: add unit tests - some are here:
 # https://github.com/mapbox/rasterio/blob/master/rasterio/mask.py
 # This function doesn't stand alone because it writes to an open object called
 # in the other function
-def stack(sources, dest):
+# @deprecate
+def _stack(sources, dest):
```
lwasser: @joemcglinchy im a bit confused here. the  deprecated function was called `stack()` we never had a `_stack`. is something duplicated in this PR by chance?

lwasser: or perhaps it just should be called stack() which i think was the original helper function name

lwasser:
@joemcglinchy  please also have a look at the travis logs

https://travis-ci.org/earthlab/earthpy/builds/481072184#L1467 
I thijnk that stack_Raster_tifs is called in the doc string of the deprecated function stack_Raster_tifs and that is causing one error.

It also looks like it can't find 
```
create_tif_file(arr, bfi)
E           NameError: name 'create_tif_file' is not defined
```

https://travis-ci.org/earthlab/earthpy/builds/481072184#L1503 

we are close on this PR!!!

joemcglinchy:
need a few more changes, as `create_raster_tif` has been blown away for the fixture `basic_image_tif`. incorporating those now

earthpy/spatial.py
joemcglinchy: stack is now the name of the 'visible' function. stack_raster_tifs used to call stack, which i changed to _stack, since it was a helper method in that context.

For the deprecated methods (stack_raster_tifs and _stack), I'm considering just blowing them away. Thoughts?

earthpy/spatial.py
mbjoseph: I think the `@deprecate` decorator is probably fine, but could see a justification for using that *and* moving deprecated things to the bottom of the module.

earthpy/spatial.py
joemcglinchy: keep it for now. when it is fully removed we can blow it away.

earthpy/spatial.py
lwasser: ok great. Thank you @mbjoseph  !!!

earthpy/spatial.py
```diff
@@ -888,3 +924,119 @@ def draw_legend(im_ax, titles=None, cmap=None, classes=None, bbox=(1.05, 1)):
...
+            # Read and return array
+            with rio.open(out_path, "r") as src:
+                return src.read(), src.profile
+        else:
+            # Write stacked gtif file
+            with rio.open(out_path, "w", **dest_kwargs) as dest:
+                return _stack(sources, dest)
+
+
+# Function to be submitted to rasterio
```
lwasser: @joemcglinchy let's remote these comments. they are dated at this point. and you have added unit tests!!

earthpy/tests/test_spatial.py
```diff
@@ -162,12 +163,80 @@ def test_bytescale_high_low_val():
...
+    if os.path.exists(out_fi):
+        os.remove(out_fi)
+
+    # Test valid use case of just getting back the array.
+    stack_arr, stack_prof = es.stack(test_files)
+
+    assert stack_arr.shape[0] == len(test_files)
+    assert stack_prof["count"] == len(test_files)
+
+    # Clean up files
```
lwasser: do we need this since it is commented out? os.remove() is commented out. or should this cleanup step happen?

earthpy/spatial.py
```diff
@@ -888,3 +924,119 @@ def draw_legend(im_ax, titles=None, cmap=None, classes=None, bbox=(1.05, 1)):
...
+        str : A path with a file name for the output stacked raster tif file.
+
+    
+    """
+
+    # Throw warning and exit
+    raise Warning("stack_raster_tifs is deprecated. Use stack(). Exiting...")
+    sys.exit()
+
+    # Set default import to read
```
lwasser: @joemcglinchy this code seems to be old and not needed. can we remove it? i suspect it's contributing to the code coverage being lower in this pr.

earthpy/spatial.py
```diff
@@ -888,3 +924,119 @@ def draw_legend(im_ax, titles=None, cmap=None, classes=None, bbox=(1.05, 1)):
...
+
+    # Throw warning and exit
+    raise Warning("_stack is deprecated. Exiting...")
+    sys.exit()
+
+    try:
+        for src in sources:
+            src.profile
+
+    except ValueError as ve:
```
lwasser: similarly this code seems extra too given the deprecation

lwasser: @mbjoseph would you agree that we shold remove deprecated code and just have a placeholder for the function so it doesn't impact our testing coverage?

earthpy/tests/test_spatial.py
joemcglinchy: i wasn't sure on this. `basic_image_tif` was never deleted anywhere in the other tests so i had it commented out. I can delete these lines.

earthpy/spatial.py
joemcglinchy: got it. done.

earthpy/spatial.py
joemcglinchy: what code... all lines of this function below line 966?

lwasser:
@joemcglinchy i'm looking at code cov and i think we don't have full coverage in your suite of tests.
here are a few items i have noticed
1. https://codecov.io/gh/earthlab/earthpy/pull/170/src/earthpy/spatial.py#L177 <- i suspect you just need a test that passes an empty list to ensure this value error is raised.
2. similarly this: https://codecov.io/gh/earthlab/earthpy/pull/170/src/earthpy/spatial.py#L251 value error is not triggered which is somehow a non  reader object is provided to the function

i'm not sure about this one

https://codecov.io/gh/earthlab/earthpy/pull/170/src/earthpy/spatial.py#L262
it seems like it hits the conditional but not the result action of the condition... maybe we can target the above first and then see what code cov says. it's possible that my comments about the deprecated functions are not impacted code coverage in this scenario. instead a few exceptions are just not hit in the tests.

earthpy/spatial.py
lwasser: yes. i guess this raises a larger question of do we KEEP old code that is deprecated and how does code cov handle that? because we don't want old code to impact our coverage stats. I did ping @mbjoseph about this as i've never deprecated a fun before. it just seems like we shouldn't maintain old code. thoughts?

earthpy/spatial.py
lwasser: ooooh i get your issue now @joemcglinchy . in this instance i think you can remove the old version of stack i get it. you have changed stack() not deprecated it. I think we just need to work about deprecating stack_Raster_tifs since that was the original parent function. is that what you mean by blowing stack away? i would still keep stack_Raster_tifs but just have the message saying it's deprecated for anyone who tries to use it. the deprecation message should just say - "hey this is deprecated - use stack() instead" . Does that make sense?

earthpy/spatial.py
joemcglinchy: removed deprecated code.

earthpy/spatial.py
joemcglinchy: removed deprecated code.

joemcglinchy:
@lwasser please review and see if this 🐶 is ready for merge!

earthpy/tests/test_spatial.py
mbjoseph: @lwasser @joemcglinchy We can remove the commented out call to `os.remove` - the `basic_image_tif` fixture does not need to be explicitly deleted because it uses the pytest `tmpdir`fixture: https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture

earthpy/tests/test_spatial.py
```diff
@@ -162,12 +163,81 @@ def test_bytescale_high_low_val():
...
             band_paths=["fname1.tif", "fname2.tif"],
             out_path="nonexistent_directory/output.tif",
         )
 
 
+def test_stack_raster(basic_image_tif):
+    """Unit tests for raster stacking with es.stack()."""
+
+    # create list of 4 basic_image_tif files (filepaths)
+    test_files = [basic_image_tif] * 4
```
mbjoseph: Since `test_files` and `band_files` are identical, do we need both @joemcglinchy? Sorry if I am missing something.

lwasser: ahhh good catch @mbjoseph i'm seeing the same thing.

earthpy/spatial.py
```diff
@@ -221,6 +241,7 @@ def stack(sources, dest):
         A list with paths to the bands you wish to stack. Objects
         will be stacked in the order provided in this list.
     dest : a rio.open writable object that will store raster data.
```
lwasser: hey @joemcglinchy  this is looking really great. i suggested a change via a commit. One other question. why do you have the dest and the write_raster flag? if we are writing a raster wouldn't you just need an ouput path to assume the user wanted to write it to the hard drive? i'm just unclear as to why we need 2 arguments here but could be missing something. i'm playing with it locally and it works great so far!!

lwasser:
ok @joemcglinchy this is SO SO close. I added one commit to this PR. i'm not sure whether you have to accept it or if that will just be automatically accepted into the  PR> essentially i updated the test in the docstring for the deprecated function. Please have a look at my comment above. I was wondering where there are two arguments for writing out a geotiff. it seems to me that you could just provide the path and if the path exists (destfile) then it would know to write a tif. Thoughts? i may just be missing something.

earthpy/spatial.py
joemcglinchy: dest and write_raster flag are both optional arguments, so the default behavior is to return an array. the flag is there as a check to make sure the user wants to write the dataset. We could instead check for a valid filename IF the user provides the filepath, but the way it is currently implemented is just an additional stage of checking that.

lwasser: ok. @joemcglinchy i think that we don't need two checks in this case. i think if the user provides a path we should 
1. check that it exists and if it exists
2. write the file 

In that instance the default value for out_path would be `None` and i'm assuming we are assuming the output driver is tiff. i could see a driver option in the future potentially i suppose.

Does that make sense to you? Or again perhaps i'm missing something here so say that if i am!!

joemcglinchy: the file won't yet exist so i think we instead check that it is valid. We'll table to driver option for now!

lwasser: right. i'm sorry i mean the path to the file not the file itself. just that you can write to that directory succesfully. Tabling the driver option sounds perfect to me!!

lwasser:
hey @joemcglinchy https://travis-ci.org/earthlab/earthpy/builds/481494480#L1466 it looks like there are a few fixes via travis now. some related to removing the write_Raster flag. Can you kindly have a look. i think we can get this merge today if we get this cleaned up.

earthpy/spatial.py
```diff
@@ -121,37 +121,31 @@ def normalized_diff(b1, b2):
...
     ----------
     band_paths : list of file paths
         A list with paths to the bands you wish to stack. Bands
         will be stacked in the order given in this list.
-    out_path : string
+    out_path : string (optional)
         A path with a file name for the output stacked raster
          tif file.
-    arr_out : boolean
+    write_raster : boolean (optional)
```
lwasser: @joemcglinchy if `write_raster` is no longer in thie function can we please remove it from the docstring? then travis is failing i think because the tests and such call it too.

joemcglinchy: yep done. write_raster is removed from test file too.

lwasser:
@joemcglinchy two tests are still failing on this PR. will you kindly look closely at the travis logs and fix them?

https://travis-ci.org/earthlab/earthpy/builds/481533971

ping me when travis is happy please.

lwasser:
just a note also that you can run pytest locally before pushing to ensure tests are running successfully!

earthpy/spatial.py
```diff
@@ -121,37 +121,28 @@ def normalized_diff(b1, b2):
...
-        stacked array and the second value
-        representing the result of src.profile of the stacked array.
-    If arr_out keyword is False:
-        str : A path with a file name for the output stacked raster tif file.
-
+    ----------
+    tuple: The first value representing the numpy array resulting from stacking the files in the input list
+        and the second value representing the result of src.profile of the stacked array.
+        NOTE: the 'count' key of the profile is updated to match the length of the input list.
+    If write_raster keyword is True:
```
lwasser: @joemcglinchy  it's here :)

lwasser:
yahoo!! thank you @joemcglinchy !! we may have set a record with all 29 commits here. I am going to MERGE THIS 🐶  now!! go team!!

Issue #166, opened 2018-12-19
Mention `pre-commit install` in contributing docs


betatim:
Via #158:

> I think it would be good to add instructions to `CONTRIBUTING.rst` for folks to execute `pre-commit install`, but we can do that after merging this...

It is a bit annoying that this isn't somehow install for the user when they clone the repo. At least that was my expectation. I'll try and find out if there is a way to do this automagically.

The fact that users could forget to do this means we should add black to the CI setup. Otherwise there will be a PR sooner or later that isn't properly formatted and we will be back to the "woah all the lines changed".

mbjoseph:
Thanks for following up with this Tim! I will assign myself to add this info the docs, and also look into running black on Travis as a check.

PR #172, opened 2018-12-19
Run black on Travis CI

mbjoseph:
This PR will run black on Travis CI to verify that the pre-commit hook has been used locally. It should fail any builds that have commits that black would change, as suggested in https://github.com/earthlab/earthpy/issues/166. 

I am going to push a commit to this branch that should break the build to verify before merging.


mbjoseph:
:raised_hands: it works!

Issue #176, opened 2018-12-20
Export all plot functions to a new module called plot


lwasser:
`import earthpy.plot as ep`  

Let's move the following functions to a new module called plot:

* draw_legend
* colorbar
* plot* plot_rgb, hist, plot_bands

to a new module called plot

*  this means we will need to update the lessons

PR #191, opened 2019-01-17
Move all plot funs to new plot module

lwasser:
This PR addresses [this issue](https://github.com/earthlab/earthpy/issues/176) and moves all plot funs to a new module. It also updates tests and docs as needed.

NOTE that we will need to update MANY of the earth analytics python lessons with this and other pr's. 

Changes here include renaming all plot and legend functions  to use the `import earthpy.plot as ep` rather than `earthpy.spatial as es`


earthpy/plot.py
```diff
@@ -0,0 +1,471 @@
...
+
+        >>> import matplotlib.pyplot as plt
+        >>> import rasterio as rio
+        >>> import earthpy.plot as ep
+        >>> from earthpy.io import path_to_example
+        >>> with rio.open(path_to_example('rmnp-dem.tif')) as src:
+        ...     dem = src.read()
+        ...     fig, ax = plt.subplots(figsize = (10, 5))
+        >>> im = ax.imshow(dem.squeeze())
+        >>> ep.colorbar(im)  #doctest: +ELLIPSIS
```
lwasser: way to tell doctest to check that a "colorbar (in this case) is returned at memor address ....

earthpy/plot.py
```diff
@@ -0,0 +1,471 @@
...
+    -------
+    matplotlib.pyplot.colorbar
+
+        Matplotlib color bar object with the correct width that matches the
+        y-axis height.
+
+    Examples
+    --------
+
+    .. plot::
```
lwasser: sphinx -- tell it that the following code chunk produces a plot. And it should be plotting it.... without it the figure won't render.

lwasser:
heya @mbjoseph i'm close to getting that stack fun merged with Joe. in the meantime i think it's safe for this PR to be reviewed whenever you have time!! travis is happy now. Codecov isn't but there are not actual code changes here so i'm not sure why or if those red x's matter. literally i just moved things around in this PR and fixed one bad test that was failing due to that `basic_image` fixture update!

lwasser:
This PR is erroring and i'm not sure why!! i've tried to rebuild a few times now. will try again later.

lwasser:
Thank you @mbjoseph !! there were a few odd things with this PR including the coverage issue with code cov that i looked into but didn't want to spend hours digging into it. it was odd since the code base was not changed. Thank you for the cleanup and merge!!

PR #200, opened 2019-01-18
Add docs prereqs

### LINKED ISSUE #198: Add dependencies for docs to Makefile
Right now `make docs` is not aware of the dependencies for the `docs` target. I need to add this to the top level Makefile so that it rebuilds the docs when it needs to.

mbjoseph:
This PR adds prerequisites to the recipe for making docs, so that the command `make docs` with execute only when necessary (when files in the prerequisites have changed). This also adds a smidgen more information to the CONTRIBUTING.rst file for how to force a rebuild.

Closes #198

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -48,10 +48,13 @@ Ready to contribute? Here's how to set up `earthpy` for local development.
 
    Now you can make your changes locally.
 
-5. When you're done making changes, check that tests pass, and examples run::
+5. When you're done making changes, check that tests pass, docs build, and examples run::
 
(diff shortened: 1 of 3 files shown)


mbjoseph:
@lwasser have a look at the changes to `Makefile` here, which will ensure that Make behaves in the way that we want (only rebuilding docs when necessary). The changes to `earthpy/tests/test_spatial.py` can be safely ignored (they were merged already)

mbjoseph:
Closed this to deal with coverage issue (forgot to merge in changes from master, which caused an apparent reduction in coverage)

PR #201, opened 2019-01-18
Add prereqs for make docs

### LINKED ISSUE #198: Add dependencies for docs to Makefile

mbjoseph:
This PR adds prerequisites to the recipe for making docs, so that the command make docs with execute only when necessary (when files in the prerequisites have changed). This also adds a smidgen more information to the CONTRIBUTING.rst file for how to force a rebuild.

@lwasser have a look at the changes to Makefile here, which will ensure that Make behaves in the way that we want (only rebuilding docs when necessary).

Closes #198

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -48,10 +48,14 @@ Ready to contribute? Here's how to set up `earthpy` for local development.
 
    Now you can make your changes locally.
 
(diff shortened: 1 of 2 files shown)


lwasser:
This looks good to me @mbjoseph !

PR #214, opened 2019-02-04
Doc string clean up for plot.py

mirob9363:
note: it appears that the commit from January 30th only comments on the black formatting done, but the majority of the changes made to this file can be found in that commit.


earthpy/plot.py
```diff
@@ -20,11 +20,12 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
-    mapobj : the matplotlib axes element.
-    size : char
-        The percent width of the colorbar relative to the plot. default = 3%
-    pad : int
-        The space between the plot and the color bar. Default = .09
+    mapobj : matplotlib axis object
+        The image that the colorbar will be representing as a matplotlib axis object.
+    size : char (optional)
+        The percent width of the colorbar relative to the plot.
+    pad : int (optional)
```
lwasser: hey @mirob9363 It's good pracice to include the default values of parameters when they exist. in this case the defaul size = 3% and the default pad = .09. Will you please add those values back? Imagine someone using this function - it's very good to see what the default values are!

lwasser: Also - it does appear as if aspect is never used in this function. We could chose to remove it from the function OR we could add it to the return statement when colorbar is called. @joemcglinchy do you have any thoughts on using the aspect argument for colorbars? do you ever use it? i haven't used it but could see it being useful.

earthpy/plot.py
```diff
@@ -70,33 +71,34 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
 
 
-# Function to plot all layers in a stack
 def plot_bands(
     arr, title=None, cmap="Greys_r", figsize=(12, 12), cols=3, extent=None
 ):
-    """Plot each layer in a raster stack read from rasterio in
-    (band, row , col) order as a numpy array. plot_bands will create an
-    individual plot for each band in a grid.
+    """Plot each layer in a raster stack read from rasterio as a numpy array.
```
lwasser: hey @mirob9363 it looks like you also removed some important information here - ie the order that the data should be in. Will you please keep this part of the docstring as it is?

earthpy/plot.py
```diff
@@ -70,33 +71,34 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
-    figsize: tuple - optional
+    arr : numpy array
+        An n-dimensional numpy array to plot.
+    title : str or list (optional)
+        Title of one band or list of titles with one title per band.
+    cmap : str (optional)
+        Colormap name for plots.
+    cols : int (optional)
+        Number of columns for plot grid.
+    figsize : tuple (optional)
```
lwasser: Same comment here - you have removed important information from the docstring. Please keep that information as it is. The cmap for example is not optional, it just has a defaule value, similarly cols is the same. please revert this to what was there before.

earthpy/plot.py
```diff
@@ -70,33 +71,34 @@ def colorbar(mapobj, size="3%", pad=0.09, aspect=20):
...
-    extent: tuple - optional
-        Bounding box that the data will fill: (minx, miny, maxx, maxy)
+    extent : tuple (optional)
+        Bounding box that the data will fill: (minx, miny, maxx, maxy).
 
     Returns
     ----------
-    fig, ax or axs : figure object, axes object
-        The figure and axes object(s) associated with the plot.
+    tuple
```
lwasser: very nice update!

earthpy/plot.py
```diff
@@ -185,30 +187,31 @@ def plot_rgb(
...
-        The extent object that matplotlib expects (left, right, bottom, top)
-    title: string (optional)
-        String representing the title of the plot
-    ax: object
-        The axes object where the ax element should be plotted. Default = none
-    figsize: tuple (optional)
-        The x and y integer dimensions of the output plot if preferred to set.
-    stretch: Boolean
-        If True a linear stretch will be applied
+    arr : numpy array
```
lwasser: Similar comment to above here. Keep default values. And note that if there is a default value the parameter is not optional in the sense it just is fixed until someone changes it. An optional parameter is one that you can use or leave out altogether!

earthpy/plot.py
```diff
@@ -376,6 +397,7 @@ def make_col_list(unique_vals, nclasses=None, cmap=None):
 
     # Create cmap list of colors
     cm = plt.cm.get_cmap(cmap)
+    print(cm)
```
lwasser: Why are you printing out something here? please remove this unless i'm missing your intent!

earthpy/plot.py
```diff
@@ -386,14 +408,16 @@ def draw_legend(im_ax, titles=None, cmap=None, classes=None, bbox=(1.05, 1)):
...
-    im : matplotlib image object created using imshow()
+    im : matplotlib image object
         This is the image returned from a call to imshow().
-    classes : list (optional)
-        A list of unique values found in the numpy array that you wish to plot.
     titles : list (optional)
         A list of a title or category for each unique value in your raster.
         This is the label that will go next to each box in your legend. If
         nothing is provided, a generic "Category x" will be populated.
+    cmap : str (optional)
```
lwasser: Similar to the above - please ensure that a parameter with a fixed value unless defined is NOT optional but something like titles which has a default set to None is optional! In this case bbox has a fixed value. With that said let's move bbox so be the SECOND parameter in this list! it's common to have parameters with default values FIRST and then optional parameters second.

mirob9363:
@lwasser I have made updates to the plot.py file based off of your comments. Please see in pull request.

lwasser:
great - thank you @mirob9363 !! reviewing this now.

Issue #204, opened 2019-01-24
Updating and renaming mask functions 


mirob9363:
I am working on some changes for the mask functions. Below are my proposed changes.  

- [x] make the first two functions `(make_cloud_mask` & `apply_cloud_mask`) private helper functions (SUNDER)
- [x] switch `vals` to an optional variable (in case the user has already created a mask they want to use 
- [x] modify code to account for if the user provides a pixel qa layer with values vs. their own mask raster 
- [x] add test for when user calls main function without `vals`
- [x] rename the functions to be clearer and more descriptive 

**I am asking for feedback on the renaming of these functions.** 

I think using the term “cloud mask” in the functions is misleading, because the functions can be used for more than just masking for clouds (i.e. snow or water). 
Some ideas for the renaming the functions are:
`make_cloud_mask` --> `_create_mask`
`apply_cloud_mask` --> `_apply_mask`
`make_apply_mask` --> `mask_array`

An aspect of the functions we also might want to modify is using the term “mask” all together to avoid confusion with rasterio’s mask. For this part, I am not quite sure what alternative term we could use, but it is something to think about! 

**Please comment your thoughts on my suggestions/other ideas you have!**

lwasser:
@joemcglinchy @mbjoseph @jlpalomino  we'd like your input on the above mask function name change that michelle is proposing. if you recall this is to apply a cloud mast to a numpy array. it creates the mask array from the pixel_qa layer if provided with values. And then applies it to the array to be masked.
```
make_cloud_mask --> _create_mask
apply_cloud_mask --> _apply_mask
make_apply_mask --> mask_array
```
Please give us some feedback on the suggested name changes above!! 
@mirob9363  please note how i tagged our team on this comment. this notifies them directly so they can see and respond. :) we are so appreciative of your help working on this package!!

joemcglinchy:
The "mask" term may be a candidate for confusion, I can see that. We have rasterio's `mask` and numpy's `MaskedArray`. I think we want to avoid calling it strictly `mask` because we typically use that for rasterio's mask, e.g., `from rasterio.mask import mask`. `mask_array` looks very close to `masked_array` from `numpy.ma`, so there could be unwanted confusion there. We could flip it and all it `array_mask`.  Without a better name, I'm okay with keeping it as `mask_array` for now.

lwasser:
[...]

Issue #194, opened 2019-01-17
Rename all helper functions to be _helper_name -- SUNDER!!! 


lwasser:
we should go through all functions and add an `_` to those and update associated functions to ensure that they are "hidden" ...

We could also consider looking into whether it appears in the docs in the future to hide these items 

DUNDER: `__fun_name__`
SUNDER: `_fun_name`

lwasser:
hey @mirob9363 @nkorinek are there any helpers left without the SUNDER added? And can one of you please add the PR's to this issue that address that! Nathan's clip PR would be included as would Michelle's WIP for the MASK function. And potentially i think plot_legend has atleast one?

lwasser:
[This PR](https://github.com/earthlab/earthpy/pull/237) addresses helpers for the mask function

nkorinek:
https://github.com/earthlab/earthpy/pull/205 addresses helpers for the clip function

mirob9363:
The changes we made to mask.py merged in https://github.com/earthlab/earthpy/pull/237 accounts for all the helper function renaming there!

lwasser:
woo hoo!! i think we've covered it all at this point SO... i will close this issue. if we have something that is missing we can always reopen. Thank you BOTH for working on this :)

PR #220, opened 2019-02-11
[WIP] Updating Mask Function 

mirob9363:
See issues #204 and #194.

I have made quite a few changes on the mask function. The area that I am having trouble with is in the main function (mask_pixels) when trying to see if the user passed in a pixel QA layer with their desired values to be masked instead of passing in a masked array they would like to use for the process. The issue I am running into is on line 211. @lwasser could you please take a look at this section, as well as my other proposed changes?


lwasser:
hey @mirob9363 it looks like you have two pull requests open now. Let's just leave this one which is a WIP to the master branch open. you don't have to contribute to the branch i made as a testing space. i will contribute back to your fork directly. i'll go ahead and close the second PR so our review doesn't get confusing :)

PR #233, opened 2019-02-15
[WIP] Update to mask and test_mask

mirob9363:
Some of the edits here are good to merge, but others I am still fixing.


lwasser:
Thank you @mirob9363 i will have a look at this -- likely on Monday when my mind is a bit more fresh!! thank you for all of the hard work on this!!

PR #237, opened 2019-02-18
Updates to mask 

lwasser:
Expecting code cov to not be 100% but wanted to see where we are.


lwasser:
ok @mirob9363 
I have cleaned up the mask tests and fun. Let's have you move on to another task next!! this seems to be ok although appveyor is failing. @mbjoseph can we ignore appveyor for the time being? We have 100% coverage here again and i believe travis will be happy  once it finishes running here.

mbjoseph:
Yes - we can ignore AppVeyor here! It is not relevant for this branch.

earthpy/mask.py
```diff
@@ -86,105 +86,107 @@
...
     Parameters
     -----------
     mask_arr : numpy array
-        An array... to open the pixel_qa or mask raster of interest
+        An array of the pixel_qa or mask raster of interest.
 
     vals : list of numbers (int or float)
-        A list of values that represent no data in the provided raster
-        layer (mask_arr)
+        A list of values from the pixel qa layer that will be used to create
```
mbjoseph:
```suggestion
        A list of values from mask_arr (the qa layer) used to create
```

mbjoseph: Referring to `mask_arr` explicitly here instead of "the pixel qa layer" might be easier for users to see the connection between the `mask_arr` and `vals` arguments

earthpy/mask.py
```diff
@@ -86,105 +86,107 @@
...
-    vals to mask.
+    pixel_qa layer for MODIS or Landsat, and apply a mask given a range of
+    values to mask. This function can also be passed a previously created
+    mask in place of a qa layer and a list of values.
 
     Parameters
     -----------
     arr : numpy array
-        An array in rasterio (band, row, col) order
+        The desired array to mask in rasterio (band, row, col) order.
```
mbjoseph:
```suggestion
        The array to mask in rasterio (band, row, col) order.
```

mbjoseph: Super minor suggestion here - but "desired" array might be confusing

earthpy/tests/test_mask.py
```diff
@@ -1,52 +1,92 @@
...
-        make_apply_mask(im, im_mask_no_ones, [3, 4])
+    with pytest.raises(
+        ValueError,
+        match="""List of values provided for the mask does 
+                             not exist in your mask array.""",
+    ):
+        mask_pixels(im, im_mask_no_ones, [3, 4])
+
+
+def test_masked_arr_provided(im, im_mask):
```
mbjoseph: This test is pretty hard to reason about because both `im` and `im_mask` are random. The two random numbers in particular (2 and 4) make it difficult to understand. For example, the line

```python
masked_im = np.ma.masked_where(im_mask < 2, im)
```

is hard to grok, because there is a chance that there are no places where `im_mask < 2`.

As far as I can see, there's no downside to having the `im` and `im_mask` fixtures be deterministic instead of random.

earthpy/mask.py
lwasser: totally!

earthpy/mask.py
lwasser: perfect!

earthpy/tests/test_mask.py
lwasser: ok @mbjoseph so you are suggesting that i hard code those two arrays rather than create one like this: `np.random.randint(10, size=(4, 5))` no problem.

lwasser:
webhooks seems to be slow today @mbjoseph  but please have another look. travis seems happy but the messages here don't show that it is finished yet it is finished running! 

https://travis-ci.org/earthlab/earthpy/pull_requests

lwasser:
just for giggles i updated frm master to see if travis talks to github when it's done!

earthpy/tests/test_mask.py
```diff
@@ -8,18 +8,14 @@
 @pytest.fixture
 def im():
     """Create image arr for masking"""
-    arr = np.array(
-        [[0, 1, 8, 7, 5], [3, 8, 8, 7, 0], [3, 0, 0, 9, 9], [5, 5, 7, 4, 5]]
-    )
+    arr = np.array([[0, 1], [2, 3]])
```
lwasser: heya @mbjoseph i don't love that this array is so simple that only one val for each potential mask element exists. I also dont love that both arrays have the same values because it just leaves room for confusion on many levels. . my thought was the mask array would have a subset of values in the larger array to be masked. can we make this a bit bigger and have larger values in the initial array.

mbjoseph: @lwasser thanks for the review! It's awesome to be able to sort out what makes for good tests together.

Is there functionality in `mask_pixels` related to the following, that is not captured in the current tests?

1. `im_mask` having a subset of values in `im`
2. Having larger arrays for `im` and `im_mask`
3. Having larger values in the `im` fixture

If so, then we should definitely test that functionality! If not, then I think there's value in having the tests be as simple as possible (so that they are easy to understand). 

If using the same array for both `im` and `im_mask` is confusing, then that's bad. I experimented with some other ideas, but it seemed easiest to go with this simple 2x2 array, because it's small and easy to understand. Do you have an alternative in mind that would be easier to grasp?

lwasser: sure. i worry that if both arrays have the same value, and there is only one instance of that value, that something could "work" by mistake. 

I'd prefer to see a slightly larger array that duplicated values to be masked so we see that it works on n+1. id also prefer that the array to do the masking has fewer values than the array to be masked. 

So maybe something like

```
arr = np.array([[0, 1, 2 ,3], [2, 3, 3, 2]])
mask_arr = np.array([[0, 1, 2, 2], [0,2, 2,1]])
```

i'd like to see an array where 
1. more than 1 value that is to be masked is masked so we know that works - that being it masks more than 1 value and
2. the mask array itself is not the same as the input array as you would expect it to be IRL.
 
In this cause i really want to see an example that represents what the data would actually be. if both masks were the same THEN you wouldn't need the second parameter at all, you'd just set all vals in the main array to NAN.

mbjoseph: Ok @lwasser  have a look at the updated fixtures, for which: 

- the fixtures are larger
- there are values in `im` that are not in `im_mask`
- there are duplicated values in `im_mask`

If we are still worried about things working by mistake, then we need test cases for those situations.

lwasser: cool. did you add a commit? i don't see it if you did.

lwasser: i see it now. github is weirdly slow today. shall we merge this not @mbjoseph  or did i somehow miss addressing other things you'd like to see? thank you!!

mbjoseph:
One last thing @lwasser - we don't currently have any tests that verify that the correct array elements are masked (so far, we have tests that ensure output is a masked array, and tests that ensure the correct number of elements are masked). 

I can add these (have some locally that I'm getting ready to commit).

lwasser:
ooooh good catch @mbjoseph you are totally correct. i didn't even think about that. it's much easier with the deterministic fixtures now. thank you.

PR #241, opened 2019-02-19
added tests for _stack_bands and deprecated stack_raster_tifs. 

joemcglinchy:
Also edited _stack_bands in spatial.py since it needed to change ValueError to Exception for except clause to be accessed. Otherwise, the specific error in the test is an AttributeError.

### DIFF
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -254,7 +254,7 @@ def _stack_bands(sources, write_raster=False, dest=None):
         for src in sources:
             src.profile
 
-    except ValueError as ve:
+    except Exception as e:
         raise ValueError("The sources object should be Dataset Reader")
         sys.exit()
 
--- a/earthpy/spatial.py
+++ b/earthpy/spatial.py
@@ -254,8 +254,8 @@ def _stack_bands(sources, write_raster=False, dest=None):
         for src in sources:
             src.profile
 
-    except Exception as e:
-        raise ValueError("The sources object should be Dataset Reader")
+    except AttributeError as ae:
+        raise AttributeError("The sources object should be Dataset Reader")
         sys.exit()
 
     else:
--- a/earthpy/tests/test__stack_bands.py
+++ b/earthpy/tests/test__stack_bands.py
@@ -1,14 +1,8 @@
 """ Tests for the _stack_bands() method """
 
 import numpy as np
-import numpy.ma as ma
-import pandas as pd
 import pytest
-import geopandas as gpd
-import rasterio as rio
-from shapely.geometry import Polygon, Point, LineString
 import earthpy.spatial as es
-import os
 
 
 @pytest.fixture
@@ -19,13 +13,13 @@ def b1_b2_arrs():
 
 
 def test__stack_bands(b1_b2_arrs):
-    """Unit test for _stack_bands()."""
+    """_stack_bands input should be of type DatasetReader."""
 
     # Test data
     b1, b2 = b1_b2_arrs
 
     # Check ValueError for Dataset Reader
     with pytest.raises(
-        ValueError, match="The sources object should be Dataset Reader"
+        AttributeError, match="The sources object should be Dataset Reader"
     ):
         es._stack_bands([b1, b2])
--- a/earthpy/tests/test__stack_bands.py
+++ b/earthpy/tests/test__stack_bands.py
@@ -1,0 +1,31 @@
+""" Tests for the _stack_bands() method """
+
+import numpy as np
+import numpy.ma as ma
+import pandas as pd
+import pytest
+import geopandas as gpd
+import rasterio as rio
+from shapely.geometry import Polygon, Point, LineString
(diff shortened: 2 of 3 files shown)


earthpy/tests/test_stack_raster_tifs.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_stack_raster_tifs(b1_b2_arrs):
+    """Unit test for stack_raster_tifs()."""
+
+    # Test data
+    b1, b2 = b1_b2_arrs
+
+    # Check Warning for Deprecation
+    with pytest.raises(Warning):
```
lwasser: @joemcglinchy  i think you want `(Warning, match="Beginning of the expected warning raised here")` to ensure the proper message is raised with the warning

earthpy/spatial.py
```diff
@@ -254,7 +254,7 @@ def _stack_bands(sources, write_raster=False, dest=None):
         for src in sources:
             src.profile
 
-    except ValueError as ve:
+    except Exception as e:
```
lwasser: @joemcglinchy i think we want to raise a value error here given that is what we are trying to catch. i might be missing something??

earthpy/tests/test__stack_bands.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.fixture
+def b1_b2_arrs():
+    b1 = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    b2 = np.array([[1, 2, 3, 4, 5], [14, 12, 13, 14, 17]])
+    return b1, b2
+
+
+def test__stack_bands(b1_b2_arrs):
+    """Unit test for _stack_bands()."""
```
lwasser: Docstring could be more specific. How about:
```
""" _stack_bands input should be of type Dataset Reader"""
```

earthpy/tests/test__stack_bands.py
```diff
@@ -0,0 +1,31 @@
+""" Tests for the _stack_bands() method """
+
+import numpy as np
```
lwasser: do we need all of these modules? I'm guessing we need
* numpy
* pytest
* rasterio
* earthpy.spatial

earthpy/tests/test_stack_raster_tifs.py
```diff
@@ -0,0 +1,29 @@
+""" Tests for the _stack_bands() method """
+
+import numpy as np
```
lwasser: Similar question - i think we can remove some of these modules!

earthpy/tests/test_stack_raster_tifs.py
```diff
@@ -0,0 +1,29 @@
...
+
+@pytest.fixture
+def b1_b2_arrs():
+    b1 = np.array([[6, 7, 8, 9, 10], [16, 17, 18, 19, 20]])
+    b2 = np.array([[1, 2, 3, 4, 5], [14, 12, 13, 14, 17]])
+    return b1, b2
+
+
+def test_stack_raster_tifs(b1_b2_arrs):
+    """Unit test for stack_raster_tifs()."""
```
lwasser: docstring update"
```
"""stack_raster_tifs is deprecated and should return a warning"""
```

earthpy/tests/test_stack_raster_tifs.py
lwasser: I think also you can also remove both comments in this test!

joemcglinchy: the pytest call did not evaluate correctly when checking for the match string, so i removed it.

earthpy/spatial.py
joemcglinchy: with my test of passing in two numpy arrays in a list as the argument to `_stack_bands`, the numpy ndarray object doesn't have a profile attribute, so an `AttributeError` is raised. I think this would be similar to any other input that is contained in a list.

earthpy/tests/test__stack_bands.py
joemcglinchy: numpy is needed for the fixture. earthpy.spatial is required for the call to `_stack_bands`. probably don't need the others though.

earthpy/spatial.py
lwasser: oh i see that. Yes so let's change them BOTH to attribute error then to be specific! what do you think?

earthpy/tests/test_stack_raster_tifs.py
lwasser: ok! i gave it a go locally and it seemed to work... i added a commit and let's see if travis yells at us or not! i've found that i don't use the ENTIRE message with match="" just the first part as spaces can throw it off. anyway let's see what happens. if it fails ill undo it!!

lwasser:
beautiful!! @joemcglinchy all tests ran, and we have 100% coverage of the spatial module now. THANK YOU for this PR!! merging :)

PR #247, opened 2019-02-20
Add URL to Rim Fire data

jlpalomino:
@lwasser  This update adds the URL to the Rim Fire data hosted on Figshare.


lwasser:
yahoo!!!  merging

PR #248, opened 2019-02-21
Add code of conduct

### LINKED ISSUE #27: Add Contributing & Code of Conduct  file s to EarthPy
Let's follow the rasterio model and add a contributing file to earthpy. I think we can start with their exact file and can adapt/ adjust it as we go.

- [ ] code of conduct file - example - https://github.com/mapbox/rasterio/blob/master/CODE_OF_CONDUCT.txt 
- [x] contributing file

mbjoseph:
This PR adds a code of conduct based on http://contributor-covenant.org and links to it from the CONTRIBUTING guidelines.

This PR also has some capitalization corrections in the docs to consistently use title case for page names. @lwasser please take a look when you get a chance and let me know if this is what you had in mind, or if you think we should make any changes to the code of conduct. 

Closes #27.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
-# EarthPy release notes
+# EarthPy Release Notes
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+* Added a code of conduct (@mbjoseph, #27)
 
 ## [0.6.2] - 2019-02-19
 We have made significant changes in preparation for a 1.0 release
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# EarthPy Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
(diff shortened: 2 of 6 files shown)


lwasser:
This looks great @mbjoseph  thank you for submitting it!
I noticed one thing. code_of_conduct using underscores in the file name vs dashes (which the other files user). I did see Will's slack post on dashes vs underscores however in this case, the file name (i think) contributes to the docs url. eg https://earthpy.readthedocs.io/en/latest/change-log.html are you adverse to using dashes in the file name vs underscores to follow web best practices?

mbjoseph:
Hi @lwasser - good question! The URL to the code of conduct will have dashes, because the rst file `code-of-conduct.rst` uses dashes. So it will be:  https://earthpy.readthedocs.io/en/latest/code-of-conduct.html

You can test this out locally if you check out the branch and `make` the docs.

lwasser:
ahhh no you are right. But it still is the only file that uses underscores! i think it's ideal to be consistent with file naming.

mbjoseph:
There is a convention to call the source file `CODE_OF_CONDUCT.md`. If we went with the hyphenated version, that's a bit less consistent with what most people do. For example, rasterio, boken, and astropy use `CODE_OF_CONDUCT.md`. There's not a huge cost here (at least that I can see) related to Will's concerns about directories with hyphenated names and regex. 

A quick comparison searching all of GitHub also suggests it's about 5x more common to use underscores: https://github.com/search?q=CODE_OF_CONDUCT.md&type=Code vs. https://github.com/search?q=CODE-OF-CONDUCT.md&type=Code

lwasser:
sure thing. Should we change the other file names then? it could be a separate PR? just to be consistent? Or is this just a convention for that particular file? I'll approve this pr but just wondering if we should be consistent across all files in the root directory.

lwasser:
Also that code cov result (lower coverage) is pretty weird!! i'm just ignoring it but it seems to be acting up this week :)

Issue #250, opened 2019-03-01
File not found error - data issue


lwasser:
hey @mbjoseph @joemcglinchy . i just installed earthpy using pip and ran into a data issue:
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-1-b7638144e60a> in <module>()
      5 import matplotlib.pyplot as plt
      6 import matplotlib.dates as mdates
----> 7 import earthpy as et
      8 import earthpy.plot as ep
      9 import geopandas as gpd

~/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/earthpy/__init__.py in <module>()
     15 
     16 epsg = json.loads(
---> 17     resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
     18 )

~/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/pkg_resources/__init__.py in resource_string(self, package_or_requirement, resource_name)
   1146         """Return specified resource as a string"""
   1147         return get_provider(package_or_requirement).get_resource_string(
-> 1148             self, resource_name
   1149         )
   1150 

~/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/pkg_resources/__init__.py in get_resource_string(self, manager, resource_name)
   1390 
   1391     def get_resource_string(self, manager, resource_name):
-> 1392         return self._get(self._fn(self.module_path, resource_name))
   1393 
   1394     def has_resource(self, resource_name):

~/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/pkg_resources/__init__.py in _get(self, path)
   1513 
   1514     def _get(self, path):
-> 1515         with open(path, 'rb') as stream:
   1516             return stream.read()
   1517 

FileNotFoundError: [Errno 2] No such file or directory: '/Users/lewa8222/anaconda3/envs/earth-analytics-python/lib/python3.6/site-packages/earthpy/example-data/epsg.json'
```

have either of you run into this? i do see that directory is NOT there in my environment.

lwasser:
this could be something specific to my push to pip as i was able to make it work installing from master

mbjoseph:
Looks like a duplicate of https://github.com/earthlab/earthpy/issues/245

Can you try the suggestion that I made in that thread to solve the issue?

lwasser:
[...]

Issue #245, opened 2019-02-20
pip install earthpy on windows 10 with Anaconda 3 does not install example-data folder


joemcglinchy:
*FEEL FREE TO DELETE IF THIS IS IDENTIFIED ALREADY OR FIXED BY AN UPCOMING RELEASE*

**Describe the bug**
pip install earthpy on windows with Anaconda 3 does not install example-data folder. This causes failure to import clip module, or earthpy as a whole.

**To Reproduce**
1. activate desired environment
2. pip install earthpy
3. run code `from earthpy import clip as cl`

**Expected behavior**
I expected the `clip` module to import without error. I also expect the `example-data` folder containing `epsg.json` to be installed.

**Screenshots**
screenshot of earthpy directory after pip install
![image](https://user-images.githubusercontent.com/4762214/53111600-94893880-34fa-11e9-9985-74a82888c196.png)

traceback upon calling `from earthpy import clip as cl`
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-1-7aec88501a42> in <module>
      3 import numpy as np
      4 from matplotlib import pyplot as plt
----> 5 from earthpy import clip as cl

C:\anaconda3\envs\eapy_gbdx\lib\site-packages\earthpy\__init__.py in <module>
     15 
     16 epsg = json.loads(
---> 17     resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
     18 )

C:\anaconda3\envs\eapy_gbdx\lib\site-packages\pkg_resources\__init__.py in resource_string(self, package_or_requirement, resource_name)
   1150         """Return specified resource as a string"""
   1151         return get_provider(package_or_requirement).get_resource_string(
-> 1152             self, resource_name
   1153         )
   1154 

C:\anaconda3\envs\eapy_gbdx\lib\site-packages\pkg_resources\__init__.py in get_resource_string(self, manager, resource_name)
   1394 
   1395     def get_resource_string(self, manager, resource_name):
-> 1396         return self._get(self._fn(self.module_path, resource_name))
   1397 
   1398     def has_resource(self, resource_name):

C:\anaconda3\envs\eapy_gbdx\lib\site-packages\pkg_resources\__init__.py in _get(self, path)
   1517 
   1518     def _get(self, path):
-> 1519         with open(path, 'rb') as stream:
   1520             return stream.read()
   1521 

[...]

PR #251, opened 2019-03-01
Remove line for including package data to fix PyPI install issue

mbjoseph:
Currently because of the way setuptools distributes data and uses the `include_package_data` argument, installing earthpy from PyPI does not seem to consistently include the example data directory as described in https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute. 

This PR solves #250 and #245, and should ensure that installing earthpy with pip includes example data.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ if __name__ == "__main__":
     setup(
         name=DISTNAME,
         maintainer=MAINTAINER,
-        include_package_data=True,
         maintainer_email=MAINTAINER_EMAIL,
         description=DESCRIPTION,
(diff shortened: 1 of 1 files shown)


lwasser:
this is great! should i try to bump a version to pip to see if it fixes things @mbjoseph ?

mbjoseph:
@lwasser if you have time could you take a look and merge this? This should fix our installation issues from PyPI.

mbjoseph:
Thanks! Yes I think we can bumpversion a patch and do a release to get this up to PyPI.

PR #252, opened 2019-03-04
added make_col_list example and fixed test issue

mirob9363:
I notice that the make_col_list example was incomplete, so I finished it.

I also noticed that the test in `test_plot_draw_legend` named `test_colors` was not passing in a list to the function. 

There is also some confusing/potentially unneeded lines of code in this test (I have not made changes to them yet though).


earthpy/tests/test_plot_draw_legend.py
```diff
@@ -115,7 +115,7 @@ def test_colors(binned_array_3bins):
     # NOTE: Do I know for sure things are rendering in the right order?
     legend_cols = [i.get_facecolor() for i in the_legend.get_patches()]
     # Get the array and cmap from axis object
-    cmap_name = im.axes.get_images()[0].get_cmap().name
+    cmap_name = list(im.axes.get_images()[0].get_cmap().name)
```
lwasser: hey @mirob9363 doesn't this call already return a list? this used to work without the call to list.

earthpy/plot.py
```diff
@@ -407,7 +407,16 @@ def make_col_list(unique_vals, nclasses=None, cmap=None):
...
     >>> import numpy as np
     >>> import earthpy.plot as ep
-
+    >>> import matplotlib.pyplot as plt
+    >>> bins = [-np.inf, 2, 7, np.inf]
+    >>> im_arr_bin = np.digitize(np.random.randint(10, size=(15, 15)), bins)
+    >>> f, ax = plt.subplots()
+    >>> im = ax.imshow(im_arr_bin, cmap="Blues")
+    >>> the_legend = ep.draw_legend(im_ax=im)
+    # Get the array and cmap from axis object
```
lwasser: Please test `make docs` and `pytest` locally before submitting prs. docs won't build here because you've added a comment without ```>>>```

earthpy/plot.py
```diff
@@ -407,7 +407,15 @@ def make_col_list(unique_vals, nclasses=None, cmap=None):
     -------
     >>> import numpy as np
     >>> import earthpy.plot as ep
-
+    >>> import matplotlib.pyplot as plt
```
lwasser: @mirob9363  i made some small changes to this to see if it fixes the build. it should

lwasser:
thank you @mirob9363 !! merging this!

PR #255, opened 2019-03-10
Test EarthlabData class

mbjoseph:
This PR adds unit tests for the `EarthlabData` class. 

**Details**

- This also adds the pytest-vcr package as a dev requirement. To adequately test the `EarthlabData` class, we need to use the internet to see whether files exist, ensure files download, etc. The pytest-vcr package ensures that these downloads only happen once, using the @pytest.mark.vcr decorator for tests. More info can be found here: https://pytest-vcr.readthedocs.io/en/latest/ and some higher level conceptual background is here: https://vcrpy.readthedocs.io/en/latest/

- This PR cleans up unused and unnecessary imports in `earthpy/__init__.py`.

- The documentation for the `EarthlabData` class has been expanded as part of this PR. There are now examples of usage, and some attempts to warn users that the default behavior is to write to `~/earth-analytics/data/`.

- This also has some improvements to error messages and removes some copypasta in places. 

- The implementation of sorting out file names from arbitrary URLs has been fixed and streamlined. 

- An example of using `path_to_example` has been added. 

- The function `list_files` has been removed, as it is not used in earthpy or in any tutorials that I could find. The glob package in the standard library provides equivalent and better tested functionality.

### DIFF
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -4,9 +4,8 @@
 
 import os
 import os.path as op
-
+import re
 import requests
-
 from download import download
 import earthpy
 
@@ -171,26 +170,12 @@ class EarthlabData(object):
             this_root = op.join(self.path, key)
 
         if url is not None:
-            # try and workout the filename and file type
-            fname = None
-            r = requests.head(url)
-            try:
-                content_disposition = r.headers["content-disposition"]
-            except KeyError:
-                raise KeyError(
-                    "content-disposition was not found in "
-                    "headers. The get_data method currently is only"
-                    "tested for figshare URLs. See the help for "
-                    "get_data for example usage."
-                )
-            for c in content_disposition.split(";"):
-                if c.startswith("filename="):
-                    fname = c.split("=")[1]
-                    break
-            else:
-                raise RuntimeError(
-                    "Could not deduce filename for " "{}.".format(url)
-                )
+            with requests.get(url) as r:
+                if "content-disposition" in r.headers.keys():
+                    content = r.headers["content-disposition"]
+                    fname = re.findall("filename=(.+)", content)[0]
+                else:
+                    fname = url.split("/")[-1]
 
             # try and deduce filetype
             file_type = "file"
@@ -198,11 +183,9 @@ class EarthlabData(object):
                 if fname.endswith(kind):
                     file_type = kind
 
-            # strip off the file extension so we get pretty download
-            # directories
+            # strip off file extension so we get pretty download directories
             if file_type != "file":
-                # cut off an extra character to remove the trailing dot as well
-                fname = fname[: -(len(file_type) + 1)]
+                fname = os.path.splitext(fname)[0]
 
             this_data = (url, fname, file_type)
             this_root = op.join(self.path, "unsorted")
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -234,44 +234,6 @@ class EarthlabData(object):
         return data_paths
 
 
-# Potential functionality for website build.
-# Move to new utils package
-
-
-def list_files(path, depth=3):
-    """
-    List files in a directory up to a specified depth.
-
-    Parameters
-    ----------
-    path : str
-        A path to a folder whose contents you want to list recursively.
-    depth : int
-        The depth of files / folders you want to list inside of ``path``.
-    """
-    if not os.path.isdir(path):
-        raise ValueError("path: {} is not a directory".format(path))
-    depth_str_base = "  "
-    if not path.endswith(os.sep):
-        path = path + os.sep
-
-    for ii, (i_path, folders, files) in enumerate(os.walk(path)):
-        folder_name = op.basename(i_path)
-        path_wo_base = i_path.replace(path, "")
-        this_depth = len(path_wo_base.split("/"))
-        if this_depth > depth:
-            continue
-
-        # Define the string for this level
-        depth_str = depth_str_base * this_depth
-        print(depth_str + folder_name)
-
-        if this_depth + 1 > depth:
-            continue
-        for ifile in files:
-            print(depth_str + depth_str_base + ifile)
-
-
 def path_to_example(dataset):
     """ Construct a file path to an example dataset.
 
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -77,17 +77,37 @@ class EarthlabData(object):
     """
     Data storage and retrieval functionality for Earthlab.
 
+    An object of this class is available upon importing earthpy as
+    ``earthpy.data`` that writes data files to the path:
+    ``~/earth-analytics/data/``.
+
     Parameters
     ----------
     path : string | None
-        The path where data is stored.
+        The path where data is stored. NOTE: this defaults to the directory
+        ``~/earth-analytics/data/``.
+
+    Examples
+    --------
+    List datasets that are available for download, using default object:
+
+        >>> import earthpy
+        >>> earthpy.data
+        Available Datasets: ['california-rim-fire', ...]
+
+    Specify a directory when instantiating a new object:
+
+        >>> import earthpy.io as eio
+        >>> data = eio.EarthlabData('.')
+        >>> data
+        Available Datasets: ['california-rim-fire', ...]
     """
 
     def __init__(self, path=None):
         if path is None:
             path = op.join(HOME, DATA_NAME)
         self.path = path
-        self.data_keys = list(DATA_URLS.keys())
+        self.data_keys = sorted(list(DATA_URLS.keys()))
 
     def __repr__(self):
         s = "Available Datasets: {}".format(self.data_keys)
@@ -109,6 +129,7 @@ class EarthlabData(object):
         url : str
             A URL to fetch into the data directory. Use this for ad-hoc dataset
             downloads. Note: ``key`` and ``url`` are mutually exclusive.
+            Currently this is only tested against figshare URLs.
         replace : bool
             Whether to replace the data for this key if it is
             already downloaded.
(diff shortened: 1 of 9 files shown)


mbjoseph:
Ok @lwasser this is ready for review!

earthpy/io.py
```diff
@@ -77,17 +76,37 @@ class EarthlabData(object):
     """
```
lwasser: I'd really like to rename this `EarthpyData()` to be consistent with the package name. I know that's not in scope for this PR but are you ok if i change it here or would you prefer a separate PR?

earthpy/io.py
```diff
@@ -117,51 +136,55 @@ def get_data(self, key=None, replace=False, url=None):
         -------
         path_data : str
             The path to the downloaded data.
+
+        Examples
+        --------
+        Download a dataset using a key:
+
+            >>> earthpy.data.get_data('california-rim-fire') # doctest: +SKIP
```
lwasser: Let's be consistent with calling this
```
import earthpy as et
et.data.get_data('california-rim-fire')
```

earthpy/io.py
```diff
@@ -117,51 +136,55 @@ def get_data(self, key=None, replace=False, url=None):
...
+            print(self.__repr__())
             return
 
         if key is not None:
             if key not in DATA_URLS:
-                raise ValueError(
-                    "Don't understand key "
+                raise KeyError(
+                    "Key " + key + " not found in earthpy.io.DATA_URLS. "
                     "{}\nChoose one of {}".format(key, DATA_URLS.keys())
```
lwasser: print this as well to render the break?

earthpy/io.py
```diff
@@ -117,51 +136,55 @@ def get_data(self, key=None, replace=False, url=None):
...
+
+        Examples
+        --------
+        Download a dataset using a key:
+
+            >>> earthpy.data.get_data('california-rim-fire') # doctest: +SKIP
+
+        Or, download a dataset using a figshare URL:
+
+            >>> url = 'https://ndownloader.figshare.com/files/12395030'
```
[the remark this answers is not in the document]
lwasser: Also for an example, here is a nice small data subset to grab!! the example here i think is like 200 mb :)
`https://ndownloader.figshare.com/articles/6890945/versions/3`

earthpy/io.py
```diff
@@ -117,51 +136,55 @@ def get_data(self, key=None, replace=False, url=None):
...
-            # strip off the file extension so we get pretty download
-            # directories
+            # strip off file extension so we get pretty download directories
             if file_type != "file":
-                # cut off an extra character to remove the trailing dot as well
-                fname = fname[: -(len(file_type) + 1)]
+                fname = os.path.splitext(fname)[0]
 
             this_data = (url, fname, file_type)
             this_root = op.join(self.path, "unsorted")
```
lwasser: i'm playing with this now but i'm not sure why an `unsorted` directory is created. Perhaps we can ask for a url and a dir_name to allow the user to specify a dirname of their choice?

earthpy/io.py
```diff
@@ -77,17 +76,37 @@ class EarthlabData(object):
...
+    --------
+    List datasets that are available for download, using default object:
+
+        >>> import earthpy
+        >>> earthpy.data
+        Available Datasets: ['california-rim-fire', ...]
+
+    Specify a directory when instantiating a new object:
+
+        >>> import earthpy.io as eio
```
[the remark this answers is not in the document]
mbjoseph: @lwasser thanks - I didn't realize we were using the convention `import earthpy as et`. So this should be `et.data.path = "."` instead of `earthpy.data.path = "."`?

earthpy/io.py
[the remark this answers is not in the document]
mbjoseph: Great catch @lwasser!

earthpy/io.py
lwasser: i tried this
1. add dir_name as an optional var in the class
then
```
if not dir_name:
    dir_name = "data-download"
this_root = op.join(self.path, dir_name)
```

earthpy/io.py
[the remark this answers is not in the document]
mbjoseph: This should be solved now - I changed `requests.head()` to `requests.get()`, which caused this issue. I just reverted this change, which should allow the `get_data` method to properly figure out the file name of this download.

earthpy/io.py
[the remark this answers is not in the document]
lwasser: oooh yes please. `et.data.path`
yes that is the convention we've been using for the past year!! but my gosh we had that discussion forever and ever ago :) thank you

earthpy/io.py
```diff
@@ -117,51 +136,55 @@ def get_data(self, key=None, replace=False, url=None):
...
                 "The `url` and `key` parameters can not both be "
                 "set at the same time."
             )
         if key is None and url is None:
-            print("Available datasets: {}".format(list(DATA_URLS.keys())))
+            print(self.__repr__())
             return
 
         if key is not None:
             if key not in DATA_URLS:
```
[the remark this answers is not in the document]
mbjoseph: Good catch @lwasser! This is all kinds of messed up. I am going to clean this KeyError to provide prettier output.

earthpy/tests/test_io.py
```diff
@@ -1,34 +1,40 @@
...
 import pytest
 import numpy as np
 import rasterio as rio
 import geopandas as gpd
-from earthpy.io import path_to_example
+import earthpy.io as eio
+
+
+@pytest.fixture
+def eld(tmpdir):
```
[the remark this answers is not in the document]
mbjoseph: I thought it was more explicit to call the class being tested by name directly in these tests, so that readers don't have to retrace where `earthpy.data` comes from (by noticing that there is not an `earthpy/data.py` module, and then discovering `earthpy.data` comes from `earthpy/__init__.py`). 

We could add `import earthpy as et`, and then reset the `earthpy.data.path` attribute, but that seemed a less direct than creating the fixture directly from the class.

earthpy/io.py
mbjoseph: I can change it in this PR - do you want `EarthpyData` over `EarthPyData`? We use the capitalization EarthPy in the documentation, and `EarthpyData` would deviate from how that convention. But, maybe `earthpy.io.EarthpyData` is redundant (two "earthpy"s)?

Another idea would be to call the class `Data`, so that we have `earthpy.io.Data`. What do you think @lwasser?

lwasser: i think the more generic it is, the better @mbjoseph  The only potential point of confusion with naming it data is that was instantiate it to `data` in the init file `data = EarthlabData()` so that would become `data = Data()`

so 
`data = Data()`
`data = EarthPyData()`
`data = AccessData()`

In theory i like doing it that way so on the front end the user doesn't have to work about instantiating an instance, they can just grab the data. Maybe then `data=Data()` is ok?

I think in general i don't expect users to call `earthpy.io.Data`  often. the most common use case will be `et.data.get_data()`

earthpy/io.py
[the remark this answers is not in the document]
lwasser: wonderful!

earthpy/io.py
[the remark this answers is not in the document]
lwasser: HAHA no worries. `.format()` is always a bit goofy to work with!

earthpy/tests/test_io.py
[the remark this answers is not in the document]
lwasser: ok sure! that makes perfect sense for the tests :) thanks!

earthpy/tests/test_plot_rgb.py
```diff
@@ -63,6 +63,7 @@ def test_1band(rgb_image):
                            order with bands first""",
     ):
         plot_rgb(a_rgb_image[1])
+    plt.close()
```
lwasser: See how weird this is! When you squash a pr it seems like sometimes when you add those commits to another PR they show up as . "changes" in the PR even tho they are in master. such a weird GH thing.

earthpy/io.py
mbjoseph: Based on our Slack convo we will change the default download directory for unsorted downloads to "earthpy-downloads/"

earthpy/io.py
mbjoseph: Sounds good - I'll change it to `earthpy.io.Data` for now, and good point about this not being super important for users, who will still use the same old `et.data`.

earthpy/tests/test_plot_rgb.py
mbjoseph: For sure - let's make sure we remember to squash merge this PR!

earthpy/io.py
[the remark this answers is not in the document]
mbjoseph: Fixed this error message with 6f279daaf765204e9249521c96b368511d80d718

earthpy/io.py
```diff
@@ -77,17 +76,37 @@ class EarthlabData(object):
...
-        The path where data is stored.
+        The path where data is stored. NOTE: this defaults to the directory
+        ``~/earth-analytics/data/``.
+
+    Examples
+    --------
+    List datasets that are available for download, using default object:
+
+        >>> import earthpy
+        >>> earthpy.data
```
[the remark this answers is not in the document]
mbjoseph: Got it - I've changed to `et.data` in the examples.

earthpy/io.py
mbjoseph: Sure - I changed the examples to use `et.data`.

earthpy/io.py
mbjoseph: Fixed with 6f279daaf765204e9249521c96b368511d80d718

earthpy/io.py
mbjoseph: Changed in 6f279daaf765204e9249521c96b368511d80d718

earthpy/io.py
mbjoseph: Changed in 9565786e7ffbcdec0507c17082a0f03a8cfcc4ed

mbjoseph:
Ok @lwasser thanks a ton for the careful review! I think this is ready for another look.

lwasser:
sure thing @mbjoseph  i'll look at it now!

lwasser:
ok @mbjoseph  this looks great!! thank you!

PR #244, opened 2019-02-20
Test across versions and platforms

mbjoseph:
This tests earthpy against python 3.5 through 3.7 using tox, and also uses AppVeyor to test earthpy on Windows. Last, this updates documentation for development using conda or virtualenv and testing of multiple python versions with tox.

As part of the review, let's take the updated development instructions in `CONTRIBUTING.rst` for a test drive. I will be curious to see what issues come up, especially on Windows. I'm opening this as a draft PR until we get some more eyes on it.

Solves the hardest part of #228 - testing on Windows (but we should leave that issue open until we get some tests in CI for macOS). Closes #213. Also closes #185.

PR #256, opened 2019-03-11
Clean up Plots in all tests

lwasser:
This PR should address the warning associated with open figures in matplotlib. Essentially `.close` closes any open windows and `.clf` clears the actual figure.

```
earthpy/tests/test_plot_rgb.py::test_stretch_output_scaled
  /Users/leahwasser/anaconda3/envs/earthpy-dev/lib/python3.6/site-packages/matplotlib/pyplot.py:514: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
    max_open_warning, RuntimeWarning)

-- Docs: https://docs.pytest.org/en/latest/warnings.html

```

### DIFF
--- a/earthpy/tests/test_plot.py
+++ b/earthpy/tests/test_plot.py
@@ -198,4 +198,4 @@ def test_hist_number_of_columns(image_array_2bands):
     for n in number_of_columns:
         f, ax = ep.hist(image_array_2bands, cols=n)
         assert [a.numCols for a in ax] == [n] * 2
-    plt.close()
+        plt.close(f)
--- a/earthpy/tests/test_plot.py
+++ b/earthpy/tests/test_plot.py
@@ -2,6 +2,9 @@
 
 import numpy as np
 import pytest
+import matplotlib as mpl
+
+mpl.use("TkAgg")
 import matplotlib.pyplot as plt
 import earthpy.plot as ep
 
@@ -15,6 +18,8 @@ def test_arr_parameter():
     """Raise an AttributeError if an array is not provided."""
     with pytest.raises(AttributeError):
         ep.plot_bands(arr=(1, 2))
+
+    plt.clf()
 
 
 def test_num_titles(image_array_2bands):
@@ -32,6 +37,7 @@ def test_num_titles(image_array_2bands):
         ep.plot_bands(
             arr=image_array_2bands, title=["Title1", "Title2", "Title3"]
         )
+    plt.clf()
 
 
 def test_num_axes(image_array_2bands):
@@ -41,6 +47,7 @@ def test_num_axes(image_array_2bands):
     """
     f, ax = ep.plot_bands(image_array_2bands)
     assert len(f.axes) == 3
+    plt.clf()
     plt.close(f)
 
 
@@ -52,6 +59,7 @@ def test_two_plot_title(image_array_2bands):
     num_plts = image_array_2bands.shape[0]
     all_titles = [ax[i].get_title() for i in range(num_plts)]
     assert all_titles == ["Band 1", "Band 2"]
+    plt.clf()
     plt.close(f)
 
 
@@ -62,6 +70,7 @@ def test_custom_plot_title(image_array_2bands):
     num_plts = image_array_2bands.shape[0]
     all_titles = [ax[i].get_title() for i in range(num_plts)]
     assert all_titles == ["Red Band", "Green Band"]
+    plt.clf()
     plt.close(f)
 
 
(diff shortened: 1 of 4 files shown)


lwasser:
ok @mbjoseph this PR closes plots throughout to avoid that plot warning issue!

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -120,6 +126,8 @@ def test_colors(binned_array_3bins):
     image_colors = ep.make_col_list(unique_vals, cmap=cmap_name)
 
     assert image_colors == legend_cols
+    plt.clf()
```
lwasser: remove

mbjoseph:
@lwasser are you sure that the calls to `plt.clf()` are necessary here? I deleted all of them and ran the tests, and I do not see the warning described in #257.

lwasser:
ok @mbjoseph i'm going to clean this up after reading this:
this states `.close` does release memory but i had read otherwise elsewhere :/ 

```
I think it is worth mentioning that plt.close() releases the memory, thus is preferred when generating and saving many figures in one run.

Using plt.clf() in such case will produce a warning after 20 plots (even if they are not going to be shown by plt.show()):
```

lwasser:
ok @mbjoseph here is the new and improved PR!! please have a look.

earthpy/tests/test_plot.py
```diff
@@ -183,3 +198,4 @@ def test_hist_number_of_columns(image_array_2bands):
     for n in number_of_columns:
         f, ax = ep.hist(image_array_2bands, cols=n)
         assert [a.numCols for a in ax] == [n] * 2
+    plt.close()
```
mbjoseph: Should we be closing the plots in the loop here instead? Calling `plt.close()` with no arguments seems to close the current plot, which is probably the last one created: https://matplotlib.org/api/_as_gen/matplotlib.pyplot.close.html

earthpy/tests/test_plot_rgb.py
```diff
@@ -154,3 +155,5 @@ def test_stretch_output_scaled(rgb_image):
         axs
     finally:
         del axs
+
+    plt.close()
```
mbjoseph: It looks like calling `plt.close()` with no arguments will only close the current figure (https://matplotlib.org/api/_as_gen/matplotlib.pyplot.close.html). Since we create 10 figures in this test, maybe this will only close the last one? 

Perhaps a test that explicitly closes each figure (e.g., https://github.com/earthlab/earthpy/pull/244/files#diff-dca3c53fc048765fa458bf9c0d489975) might be better, because we can be sure each figure is closed.

earthpy/tests/test_plot.py
lwasser: oops yes!

earthpy/tests/test_plot_rgb.py
lwasser: just curious - how are you calling an ax object after closing a plot?

mbjoseph: I'm not sure I understand the question - the ax object is appended to a list `axs` before closing the plot in the above link.

earthpy/tests/test_plot_rgb.py
```diff
@@ -146,11 +147,15 @@ def test_stretch_output_scaled(rgb_image):
...
-    assert n_unique_means == len(stretch_vals)
+
+    mean_vals = list()
+    for v in stretch_vals:
+        ax = plot_rgb(arr, stretch=True, str_clip=v)[1]
+        mean = ax.get_images()[0].get_array().mean()
+        mean_vals.append(mean)
+        plt.close()
+    assert len(set(mean_vals)) == len(stretch_vals)
     try:
```
mbjoseph: @lwasser this looks better! Sorry if I'm missing something, but what is this try/finally logic doing at the end of this test? Can we get rid of it? I am not seeing how it's relevant to the assertion that we test against.

lwasser: it's deleting the ax object which remains open post figure close. it's super confusing to me why the ax obj is still open after you close the figure. i guess `plt.cla()` would clear that? does it matter @mbjoseph

mbjoseph: Because of the way that pytest isolates test environments, the `ax` object and any other objects created in the test should be deleted after the test is finished. 

I just tested locally with deleting the try/except statement, and all the tests still pass with no warning, so I think we should go ahead and delete it @lwasser.

lwasser: ok great. yes it doesn't break anything. it just cleans up ALL elements of the plot object. i'll remove it.

lwasser:
ok @mbjoseph  please give this another review!

mbjoseph:
Looks good @lwasser!

Issue #259, opened 2019-03-12
Do we need earthpy.utils.fix_paths?


mbjoseph:
The `fix_paths` function in `earthpy/utils.py` seems like a carry-over from website build logic. Can we remove this from EarthPy?

lwasser:
oh totally. that is a no brainer. i'll submit a PR now!!

lwasser:
Closing this as the PR was just merged!! Thank you for creating this issue @mbjoseph !!!

Issue #199, opened 2019-01-18
Remove deprecated code from earthpy


lwasser:
In the next pr's remove the `deprecated.py` script from earthpy as it has a suite of syntax issues and is throwing off the docs build.

- [ ] Thing two 
- [x] `fix_paths()`

PR #262, opened 2019-03-12
remove utils.py as it's not being used in earthpy

lwasser:
Addressing 
* https://github.com/earthlab/earthpy/issues/259 
* https://github.com/earthlab/earthpy/issues/199


lwasser:
@mbjoseph wanna check this puppy out? :)

PR #263, opened 2019-03-12
Bug in plot_bands with single titles

lwasser:
@mirob9363  discovered a bug in `plot_bands` where it failed because I was calculated the number of titles using len which in a string calculates characters! To account for this, i've added a test that tests for a string provided to the function. I've also turned the title into a list to ensure its length is calculated properly.

### DIFF
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -129,7 +129,7 @@ def plot_bands(
         # A 2-dim array should only be passed one title
         if arr.ndim == 2 and len(title) > 1:
             raise ValueError(
-                "plot_bands expects one title for a single"
+                "plot_bands expects one title for a single "
                 "band array. You have provided more than one title."
             )
         # A 3 dim array should have the same number of titles as dims
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -521,7 +521,7 @@ def draw_legend(im_ax, bbox=(1.05, 1), titles=None, cmap=None, classes=None):
 
     if not len(classes) == len(titles):
         raise ValueError(
-            "The number of classes should equal the number of"
+            "The number of classes should equal the number of "
             "titles. You have provided {0} classes and {1} titles.".format(
                 len(classes), len(titles)
             )
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -127,7 +127,7 @@ def plot_bands(
             title = [title]
 
         # A 2-dim array should only be passed one title
-        if (arr.ndim == 2) and (len(title) > 1):
+        if arr.ndim == 2 and len(title) > 1:
             raise ValueError(
                 """Plot_bands() expects one title for a single
                              band array. You have provided more than one
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -479,7 +479,7 @@ def draw_legend(im_ax, bbox=(1.05, 1), titles=None, cmap=None, classes=None):
     except AttributeError:
         raise AttributeError(
             "The legend function requires a matplotlib axis object to "
-            " run properly. You have provided a {}.".format(type(im_ax))
+            "run properly. You have provided a {}.".format(type(im_ax))
         )
 
     # If classes not provided, get them from the im array in the ax object
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -351,7 +351,7 @@ def hist(
         n_layers = arr.shape[0]
         if title and not len(title) == n_layers:
             raise ValueError(
-                "The number of plot titles should be the"
+                "The number of plot titles should be the "
                 "same as the number of raster layers in your array."
             )
         # Calculate the total rows that will be required to plot each band
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -262,7 +262,7 @@ def plot_rgb(
 
     if len(arr.shape) != 3:
         raise ValueError(
-            "Input needs to be 3 dimensions and in rasterio"
+            "Input needs to be 3 dimensions and in rasterio "
             "order with bands first"
         )
 
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -135,7 +135,7 @@ def plot_bands(
             )
         # A 3 dim array should have the same number of titles as dims
         if arr.ndim > 2:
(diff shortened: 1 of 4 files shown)


lwasser:
hey there @mbjoseph  we found a bug in `plot_bands` and this pr attempts to fix it. Can you kindly have a look whenever you have time?

earthpy/plot.py
```diff
@@ -123,17 +123,23 @@ def plot_bands(
         "Input arr should be a numpy array"
 
     if title:
+        if isinstance(title, str):
+            title = [title]
+
+        # A 2-dim array should only be passed one title
         if (arr.ndim == 2) and (len(title) > 1):
```
mbjoseph: We could simplify this line to 
```suggestion
        if arr.ndim == 2 and len(title) > 1:
```

earthpy/plot.py
```diff
@@ -123,17 +123,23 @@ def plot_bands(
...
             raise ValueError(
                 """Plot_bands() expects one title for a single
```
mbjoseph: Two minor suggestions to improve these error messages: 

1. We should probably use `plot_bands` instead of `Plot_bands`, even though it is the first word in a sentence, so that the error message has the same capitalization as the function. 

2. Tim snuck a handy trick for multiline error messages in the io module that is a little cleaner and does not result in the awkward tabs that come with the use of `"""` with indendation. See https://github.com/earthlab/earthpy/blob/master/earthpy/io.py#L163 for an example, but the idea is to do something like:

```python
raise ValueError(
  "plot_bands expects..."
  "and the error message continues..."
)
```

earthpy/plot.py
```diff
@@ -123,17 +123,23 @@ def plot_bands(
...
                              band array. You have provided more than one
                              title."""
             )
-        elif not (len(title) == arr.shape[0]):
-            raise ValueError(
-                """Plot_bands() expects the number of plot titles
+        # A 3 dim array should have the same number of titles as dims
+        if arr.ndim > 2:
+            if not (len(title) == arr.shape[0]):
```
mbjoseph:
```suggestion
            if len(title) != arr.shape[0]:
```

earthpy/plot.py
```diff
@@ -123,17 +123,23 @@ def plot_bands(
...
+                raise ValueError(
+                    """Plot_bands() expects the number of plot titles
```
mbjoseph: Same suggestion here about `plot_bands` and using `"` instead of `"""`

lwasser:
OMGoodness @mbjoseph  BONUS SUGGESTIONS?! No-You-DiD-N-T.
On it. :)

earthpy/plot.py
lwasser: @mbjoseph  this is great. so... in the past i've had issues with the multi line use of single `"` but i'm happy to try again.
Just a note that we have `"""` EVERYWHERE in earthpy. So if we change it here we might think about... changing it EVERYWHERE!! thoughts?

lwasser:
ok @mbjoseph  i went through and made the changes you suggested. i didn't update ALL of the error messages with `"""` throughout earthpy but will make an issue to do it now.

earthpy/plot.py
```diff
@@ -120,20 +120,25 @@ def plot_bands(
...
-        elif not (len(title) == arr.shape[0]):
+        if isinstance(title, str):
+            title = [title]
+
+        # A 2-dim array should only be passed one title
+        if arr.ndim == 2 and len(title) > 1:
             raise ValueError(
-                """Plot_bands() expects the number of plot titles
-                             to equal the number of array raster layers."""
+                "plot_bands expects one title for a single"
```
mbjoseph: This will print as `...one title for a singleband array` - should it be `single band` instead of `singleband`?
```suggestion
                "plot_bands expects one title for a single "
```

earthpy/plot.py
```diff
@@ -257,8 +262,8 @@ def plot_rgb(
 
     if len(arr.shape) != 3:
         raise ValueError(
-            """Input needs to be 3 dimensions and in rasterio
-                           order with bands first"""
+            "Input needs to be 3 dimensions and in rasterio"
```
mbjoseph:
```suggestion
            "Input needs to be 3 dimensions and in rasterio "
```

earthpy/plot.py
```diff
@@ -346,9 +351,8 @@ def hist(
         n_layers = arr.shape[0]
         if title and not len(title) == n_layers:
             raise ValueError(
-                """"The number of plot titles should be the
-                    same as the number of raster layers in
-                    your array."""
+                "The number of plot titles should be the"
```
mbjoseph:
```suggestion
                "The number of plot titles should be the "
```

earthpy/plot.py
```diff
@@ -519,9 +521,8 @@ def draw_legend(im_ax, bbox=(1.05, 1), titles=None, cmap=None, classes=None):
 
     if not len(classes) == len(titles):
         raise ValueError(
-            """The number of classes should equal the number of
-                                 titles. You have provided {0} classes and {1}
-                                 titles.""".format(
+            "The number of classes should equal the number of"
```
mbjoseph:
```suggestion
            "The number of classes should equal the number of "
```

earthpy/plot.py
mbjoseph: I think just changing it here for now is good!

lwasser:
ok @mbjoseph a few more edits here!! have a look

earthpy/plot.py
```diff
@@ -477,11 +478,8 @@ def draw_legend(im_ax, bbox=(1.05, 1), titles=None, cmap=None, classes=None):
         im_ax.axes
     except AttributeError:
         raise AttributeError(
-            """Oops. The legend function requires a matplotlib
-                         axis object to run properly. You have provided
-                         a {}.""".format(
-                type(im_ax)
-            )
+            "The legend function requires a matplotlib axis object to "
+            " run properly. You have provided a {}.".format(type(im_ax))
```
mbjoseph:
```suggestion
            "run properly. You have provided a {}.".format(type(im_ax))
```

lwasser:
woo hoo!! thank you @mbjoseph merging this now and then updating my other branch so the edits are in sync.

Issue #249, opened 2019-02-26
Make earthpy installable from conda-forge


mbjoseph:
EarthPy would be easier for users to install if it was on conda-forge, e.g., if they could do: `conda install -c conda-forge earthpy`. Instructions are here for adding a package to conda-forge: https://conda-forge.org/#add_recipe

One challenge we face currently is the dependency on the [download](https://github.com/choldgraf/download) package - which is not on conda-forge. To get EarthPy on conda-forge, we would either need to get the download package on conda-forge first (https://github.com/conda-forge/staged-recipes#6-do-all-of-my-packages-dependencies-have-to-be-in-conda-forge-already), or remove the dependency on the download package. 

Given that we only use the download package once, to download files, it seems like it would be fairly easy to remove it as a dependency. 

What do you think @lwasser? Would you be comfortable removing the dependency on download, using something from the standard library like requests or urllib instead?

lwasser:
Download is used in the `get_data` method. it's a small package that @choldgraf wrote. We could ask him if he's open to moving it to conda-forge. Or we could find an alternative. What are your thoughts on alternatives for downloading? Or Chris - have you considered porting this to conda-forge?

mbjoseph:
@lwasser there are a few alternatives that we could use to download files (e.g., described [here](https://stackabuse.com/download-files-with-python/)). 

But, if @choldgraf is interested in putting the download package on conda-forge, that would be great. I am happy to help with this process too.

choldgraf:
I unfortunately have no idea how to put something on conda-forge :-(

mbjoseph:
Conda-forge has some instructions here: https://conda-forge.org/#add_recipe

If it was of interest, I could work on submitting the download library to conda-forge and add you as a collaborator on the resulting repo/keep you in the loop on the submission PR. 

Otherwise, we may need to drop the download library as a dependency to get EarthPy deployed to conda-forge. 

What do you think @choldgraf - do you have any interest in getting the download library on conda-forge, and do you want me to help with the submission process?

choldgraf:
one must be possible to specify dependencies that are only installable via `pip` in a conda forge project. People do all kinds of crazy stuff with conda forge, and since conda also ships with pip, it must be possible to say "when you run `conda install earthpy` also run `pip install download`". Did you double check this?

Otherwise, I am happy to try figuring this out

mbjoseph:
[...]

PR #264, opened 2019-03-14
Remove dependency on download library

mbjoseph:
This reimplements data downloads for the `earthpy.io.Data` class, removing the dependency on the download library. This will help with #249.

There's one minor change to the behavior of `earthpy.io.Data` with this PR: instead of automatically extracting `.tar` and `.tar.gz` files, this will just download them. @lwasser if this is a problem for any of the lessons, let me know! I couldn't find any instances where there were tar files being used. With this implementation, users would download the files, then untar them afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Remove dependency on download library (@mbjoseph #249)
 * Remove earthpy.utils.fix_paths as it is not used in the package (@lwasser #259)
 * Adding tests for hillshade and improved docs (@jpalomino #260)
 * Closing plots in tests (@lwasser #257)
-* Added a code of conduct (@mbjoseph, #27)
-* Added tests for EarthlabData class (@mbjoseph, #37)
+* Added a code of conduct (@mbjoseph #27)
+* Added tests for EarthlabData class (@mbjoseph #37)
 
 ## [0.6.2] - 2019-02-19
 We have made significant changes in preparation for a 1.0 release
--- a/docs/earthpy.rst
+++ b/docs/earthpy.rst
@@ -44,14 +44,6 @@ earthpy.spatial module
     :undoc-members:
     :show-inheritance:
 
-earthpy.utils module
---------------------
-
-.. automodule:: earthpy.utils
-    :members:
-    :undoc-members:
-    :show-inheritance:
-
 
 Module contents
 ---------------
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -1,11 +1,12 @@
 """File Input/Output utilities."""
 
 
+import io
 import os
 import os.path as op
 import re
 import requests
-from download import download
+import zipfile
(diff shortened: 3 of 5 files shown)


lwasser:
hey @mbjoseph  we added support for `tar.gz` previously because landsat comes down that way.. While it is true that i've placed some files online as .zip files,  I'd like to maintain that support given i do plan to add a bit more data to my course this semester. We have figshare repos that do have tar.gz files as well.. Students  will need to bring those files into the jupyter hub for example and it would be nice to ensure that additional steps are not necessary which is why we initially added that support. is it difficult to add it back?

mbjoseph:
Ok thanks @lwasser! I will revise this to automatically extract `tar.gz` and `tar` files.

mbjoseph:
Alright @lwasser I added support for `.tar` and `.tar.gz` files, along with tests to make sure that these work as expected. This PR should be ready for review!

lwasser:
great! thanks @mbjoseph i will have a look at this pr as well.

mbjoseph:
Thanks @lwasser! I didn't realize we needed support for extracting `.gz` files as well. I will add this. 

Do you have a preference for a message vs. progress bar? Messages are easy enough (e.g., "Downloading this file..."), and if we want progress bars, we can add back the dependency on tqdm.

lwasser:
@mbjoseph you know, i'm not particular on message vs progress bar as long as the user knows something is happening
whatever is easiest i think is best. would the message blink or something so they knew something was happening?

mbjoseph:
@lwasser I'm working on revising this PR, but noticed that earthpy (using the download library, on the master branch) currently does not extract `.gz` files such as the one you linked to: https://ndownloader.figshare.com/files/14630150

I would suggest that we open a separate issue for such files because it's adding a new feature, but this PR is intended to remove a dependency while not changing the behavior or losing/adding features. Would you be comfortable opening a new issue for a feature request to add automatic decompression of binary `.gz` files such as the `.tif.gz` file in the above URL?

lwasser:
hey @mbjoseph  that works for me. i just opened [this issue about it.](https://github.com/earthlab/earthpy/pull/264#issuecomment-475095319)

mbjoseph:
Great - thanks @lwasser. This PR is ready for review. I added a `verbose` argument that defaults to `True` and prints messages to let users know that files are being downloaded.

lwasser:
hey @mbjoseph ! this looks good. i just tested and all things seem to work well! I am not noticing any changes required but do have a question. it looks like we are ingesting some code from the download library to remove ourselves from that dependency. I am ok with this however it does mean more code for us to curate / keep tabs on. I suppose that is ok for the time being until chris perhaps moves download to conda-forge?

mbjoseph:
Thanks @lwasser - the function arguments and the API are borrowed from the download library but the implementation is simpler for this code. It's definitely a trade-off between rolling some of our own functionality vs. relying on another library. If the download library does get a conda-forge deployment, we can definitely go back to using it, but given the environment issues that we've been running into elsewhere, I think we can justify incorporating this download logic to expedite the deployment of earthpy on conda-forge.

Issue #154, opened 2018-12-12
epsg lookup is undocumented - should we include this or use something from proj4?


lwasser:
currently we have an epsg dictionary that allows you to lookup the prj string for an epsg code rather than having to google it! this gives you units and such. 

We need to add epsg to the documentation of earthpy. 

et.epsg["epsg-code-here"] 

This issue is probably do we want to keep or **remove this** from our package??!!  @joemcglinchy can you please look into this?? ... it could be a function that uses `proj4` to get the string??

joemcglinchy:
at this point I would say let us leave it in earthpy! I have done some digging and have struggled at how to produce the same string we have in the `epsg.json` dictionary. For example, the first entry:

"29188": "+proj=utm +zone=18 +south +ellps=aust_SA +towgs84=-66.87,4.37,-38.52,0,0,0,0 +units=m +no_defs"

I can run 
```
from pyproj import Proj
p = Proj(init='epsg:29188')
print(p.srs)
--> '+units=m +init=epsg:29188 '
```
The closest I can get is by calling `.to_latlong().srs`, which isn't even correct.
```
print(p.to_latlong().srs)
--> b' +proj=latlong +ellps=aust_SA +towgs84=-66.87,4.37,-38.52,0,0,0,0'
```

I know that our `epsg.json` dictionary is coming from a file in the PyProj package, but I'm not seeing how  those strings are directly callable from a Proj object. My vote is to keep the dictionary and periodically check for updates from PyProj, unless someone has a better understanding of how to produce that string using the PyProj library.

lwasser:
if we keep this we will need to add documentation!! Call it in the lessons !! @nkorinek

mbjoseph:
We can document this by wrapping the existing `epsg` object in a function with a docstring. The wrapper will return what is currently used to define `epsg`. 

See `earthpy/__init__.py` for where this should happen.

lwasser:
closing this via the above merged PR!

PR #268, opened 2019-03-14
Add epsg documentation

joemcglinchy:
…g=... line

### DIFF
--- a/earthpy/__init__.py
+++ b/earthpy/__init__.py
@@ -18,9 +18,8 @@ def load_epsg():
     Example
     -------
     >>> import earthpy as et
-    >>> epsg = et.load_epsg()
     >>> # Get Proj4 string for EPSG code 4326 
-    >>> epsg4326_proj4 = epsg['4326']  
+    >>> et.epsg['4326']  
     '+proj=longlat +datum=WGS84 +no_defs'
     """
 
--- a/earthpy/__init__.py
+++ b/earthpy/__init__.py
@@ -10,3 +10,11 @@ data = Data()
 epsg = json.loads(
     resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
 )
+
+
+def load_epsg():
+    """ A function to return a dictionary of EPSG code to Proj4 string mappings."""
+
+    return json.loads(
+        resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
+    )
--- a/earthpy/__init__.py
+++ b/earthpy/__init__.py
@@ -5,13 +5,6 @@ import json
 from .io import Data
 
 
-data = Data()
-
-epsg = json.loads(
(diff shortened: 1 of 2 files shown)


earthpy/__init__.py
```diff
@@ -10,3 +10,11 @@
 epsg = json.loads(
```
lwasser: i think this part will go away since you have this in the function below now.

earthpy/__init__.py
```diff
@@ -10,3 +10,11 @@
...
     resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
 )
+
+
+def load_epsg():
+    """ A function to return a dictionary of EPSG code to Proj4 string mappings."""
+
+    return json.loads(
+        resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
+    )
```
lwasser: SO i think AFTER  the function you can define epsg like this:

`epsg = load_epsg()`

And by doing this the existing behavior is mimicked. 

FINALLY -- you will need a test to ensure epsg works. I think what you could do is just call `epsg["some-code-here"]` and assert that the return is the proj4 that you expect.

earthpy/__init__.py
```diff
@@ -10,3 +10,11 @@
 epsg = json.loads(
     resource_string("earthpy", "example-data/epsg.json").decode("utf-8")
 )
+
+
+def load_epsg():
```
lwasser: @mbjoseph  this works well BUT we need to document 

epsg["code"] returns ... which isn't explicitly built into the function but kind of is. SHould we just do that here and add some example "vignette" code to the function with how to call it?

so

```
>>> epsg["4326"]
expected return here
```

lwasser: @joemcglinchy  i just chatted with @mbjoseph 
we could do this

https://stackoverflow.com/questions/20227051/how-to-document-a-module-constant-in-python/20230473#20230473 
and just try `""" """` below the call to epsg. this would mean

1. ignore my first comment about removing the code above and 
2. remove this function and then
3. document the epsg element below the call to define it above.

let me know if that is clear as mud!

joemcglinchy:
OK, updated the `__init__.py` file and added a test file for it. We are now @ 100 tests, all passing. Hope this passes your review @lwasser !!

lwasser:
Thank you @joemcglinchy  ! rasterio is broken today and i made the mistake of trying to reinstall! SO.. i won't be able to check this until I can get things installed :) ... more soon!

lwasser:
Thank you @joemcglinchy i've made a few changes to simplify things a bit!! i'll merge this in a few if it passes travis tests! it should :) this is great!

lwasser:
ok merging this for now @joemcglinchy  thank you!!

lwasser:
this addresses [this issue](https://github.com/earthlab/earthpy/issues/154)

Issue #253, opened 2019-03-06
plot.draw_legend fails when classes provided


mirob9363:
**Describe the bug**
The `draw_legend` function is not allowing classes to be passed in. 

**To Reproduce**
Example can be found in Earth Analytics lesson:
https://github.com/earthlab/earth-analytics-lessons/blob/master/courses/earth-analytics-python/08-multispectral-remote-sensing-fire/in-class/2017-03-01-fire02-calculate-NBR-with-landsat-Python.ipynb

**Expected behavior**
Use input classes for legend

**Screenshots**

```
---------------------------------------------------------------------------`
ValueError                                Traceback (most recent call last)
<ipython-input-18-813df5b17d83> in <module>()
     14 ep.draw_legend(im, 
     15                classes=values,
---> 16                titles=dnbr_cat_names)
     17 
     18 

~\AppData\Local\Continuum\anaconda3\envs\earth-analytics-python\lib\site-packages\earthpy\plot.py in draw_legend(im_ax, bbox, titles, cmap, classes)
    479     # If classes not provided, get them from the im array in the ax object
    480     # Else use provided vals
--> 481     if classes:
    482         try:
    483             # Get the colormap from the mpl object
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
**What Operating System Are you Running? **
 - Windows

mirob9363:
@lwasser Can you please look at this?

lwasser:
@mirob9363  thank you for this PR!! can you kindly provide a link to a notebook where you IMPLEMENTED the actual `draw_legend()` functionality and it failed so i can look at that? This might mean linking to the notebook on a branch you are working on! 

Thank you

lwasser:
@mirob9363 Also this notebook uses old deprecated earthpy functions. So please do the following

1. update the notebook so that it is using the most current code.
2. Post a link to that and ideally pull out the code that is not running properly and also what error you are getting

Ping me when this is done!

mirob9363:
This example should work (or not work I guess) to demonstrate the issue: 
```
import numpy as np
import earthpy.plot as ep
import matplotlib.pyplot as plt
im_arr = np.random.uniform(-2, 1, (15, 15))
bins = [-np.Inf, -0.8, 0.8, np.Inf]
im_arr_bin = np.digitize(im_arr, bins)
cat_names = ["Class 1", "Class 2", "Class 3"]
values = np.unique(im_arr_bin.ravel())
[...]

PR #271, opened 2019-03-14
Fix bug in draw_legend where classes provide as a numpy array breaks the function

lwasser:
this pr addresses [this issue](https://github.com/earthlab/earthpy/issues/253) submitted by @mirob9363 . Essentially when the classes parameter of `draw_legend` is provided with a numpy array rather than a list (which is likely given the use of `np.unique()` on arrays, the function fails. Here i've provided a fix for this issue and also added a test that provides the function with an array to ensure it still functions properly.


lwasser:
@mbjoseph  i've made a small edit to the draw_legend function to address a big that @mirob9363 found. can you have a look at this PR? i've been careful to not mess with error messages in this pr given my other PR addresses them!

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -174,6 +174,21 @@ def test_noncont_listed_cmap(binned_array, listed_cmap):
...
 
+def test_classes_provided_as_array(binned_array, listed_cmap):
+    """Test legend fun to ensure classes provided as an array still work."""
+
+    cmap, norm = listed_cmap
+    bins, arr_class = binned_array
+
+    f, ax = plt.subplots(figsize=(5, 5))
+    im = ax.imshow(arr_class, cmap=cmap, norm=norm)
+    leg = ep.draw_legend(im, classes=np.asarray([1, 2, 3, 4, 5]))
```
mbjoseph: Minor quibble here, but we have the list explicitly written out as `[1, 2, 3, 4, 5]` twice in this test. It might be cleaner to do something like 

```python
    ...
    im = ax.imshow(arr_class, cmap=cmap, norm=norm)
    n_classes = 5
    leg = ep.draw_legend(im, classes=np.arange(n_classes))

    legend_cols = [i.get_facecolor() for i in leg.get_patches()]
    assert len(legend_cols) == n_classes
    plt.close(f)
```

lwasser: ok @mbjoseph  . i had to tweak a few things for that to work given it's 0-based indexing and such! see what you think.

mbjoseph: Cool - I am noticing that the test still passes even when doing `np.arange(n_classes)`. Is that not supposed to work?

lwasser: it should work. So essentially for a listed colormap you need to provide the number of classes and it will create that number of legend items. They are discrete values so 5 classes = 5 colors. However what should fail is if they provide 6 classes and a listed colormap of 5 colors. in that case the last two values will be the same color.  

the function is tricky because drawing a legend is a manual task in this instance - particularly when some values may be missing and only the user knows what values those are. so i tried to make it very flexible. 

i may add another test for if a use provides 6 classes but there are only 5 colors in the cmap.

mbjoseph: Ok sounds good @lwasser. I'd advocate for making the tests only as complex as they need to be to test behavior, so in this case using `np.arange(n_classes)` instead of `np.arange(1, 1 + n_classes)` would be the preferred option. 

Let me know when you're ready for another review! (maybe after adding that additional test?)

lwasser: will do. @mbjoseph what do you thijnk about this... so in this test if i rewrite it i could test 2 things.

1. that an array works and 
2. that i get a nice message (i need to add that) if a user provides more classes than colors. 

the code would be the same. Would it be better to repeat the code and have 2 tests OR to have 1 more concise test that hits both things.

mbjoseph: I would try to shoot for two tests (each with one `assert` statement), since these seem like two different things to test. 

To avoid copypasta, it might be good to think about whether there is a way to use a fixture to share some of the logic among tests. I'm noticing that there already is a fair bit of repetition and shared logic among tests. Maybe some of that can be bundled up in a fixture?

lwasser: ok i'm working on the second test FIRST because it also means adding a test into the function. then i'll think about fixtures! thanks!

lwasser:
ok @mbjoseph i've done a lot of test cleanup here. it turns out that some of these tests weren't even correct but they were magically passing. Please have a look when you get a chance.

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -30,169 +30,238 @@ def binned_array_3bins():
...
 
     f, ax = plt.subplots(figsize=(5, 5))
-    im_ax = ax.imshow(im_arr_bin, cmap="Blues")
+    return ax.imshow(arr_class, cmap=cmap), binned_array[1]
 
-    with pytest.raises(ValueError):
+
+@pytest.fixture
+def vals_missing_plot_list_cmap(binned_array, listed_cmap):
+
```
mbjoseph: A docstring here would be useful to explain what this fixture is

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -30,169 +30,238 @@ def binned_array_3bins():
...
+    arr[arr == 1] = 2
+    arr[arr == 5] = 4
+
+    f, ax = plt.subplots(figsize=(5, 5))
+    return ax.imshow(arr, cmap=cmap, norm=norm), arr
+
+
+@pytest.fixture
+def vals_missing_plot_cont_cmap(binned_array):
+
```
mbjoseph: Docstring here would also be useful.

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -30,169 +30,238 @@ def binned_array_3bins():
...
     image has classes 2,3,4 and the user wants classes 1-5 to appear """
 
-    cmap, norm = listed_cmap
-    bins, arr_class = binned_array
+    im_ax, arr = vals_missing_plot_list_cmap
 
-    f, ax = plt.subplots(figsize=(5, 5))
-    im = ax.imshow(arr_class, cmap=cmap, norm=norm)
-    leg = ep.draw_legend(im, classes=[1, 2, 3, 4, 5])
+    leg = ep.draw_legend(im_ax, classes=[1, 2, 3, 4, 5])
```
mbjoseph: Instead of writing out `[1, 2, 3, 4, 5]` twice in this test, you could define that as an object, e.g., `class_list` below: 

```python
def test_listed_cmap_3_classes(vals_missing_plot_list_cmap):
    """Test legend for a listed cmap where
    the user wants all classes to be drawn in the legend. IE the classified
    image has classes 2,3,4 and the user wants classes 1-5 to appear """

    im_ax, arr = vals_missing_plot_list_cmap

    class_list = list(range(5))
    leg = ep.draw_legend(im_ax, classes=class_list)
    legend_cols = [i.get_facecolor() for i in leg.get_patches()]

    assert len(legend_cols) == len(class_list)
    plt.close()

```

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -30,169 +30,238 @@ def binned_array_3bins():
...
+
+
+def test_cont_cmap_3_classes(vals_missing_plot_cont_cmap):
+    """Test legend for a listed cmap where
+    the user wants all classes to be drawn in the legend. IE the classified
+    image has classes 2,3,4 and the user wants classes 1-5 to appear """
+
+    im_ax, arr = vals_missing_plot_cont_cmap
 
+    leg = ep.draw_legend(im_ax, classes=[1, 2, 3, 4, 5])
```
mbjoseph: Similar change could be made here to avoid the redundant `[1, 2, 3, 4, 5]`.

earthpy/tests/test_plot_draw_legend.py
```diff
@@ -30,169 +30,238 @@ def binned_array_3bins():
...
+@pytest.fixture
+def arr_plot_list_cmap(binned_array, listed_cmap):
+    """Returns a plot rendered using a listed cmap """
+
+    cmap, norm = listed_cmap
+    bins, arr_class = binned_array
 
     f, ax = plt.subplots(figsize=(5, 5))
-    im_ax = ax.imshow(im_arr_bin, cmap="Blues")
+    return ax.imshow(arr_class, cmap=cmap), binned_array[1]
```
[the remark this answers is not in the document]
lwasser: definitely a bit cleaner this way

lwasser:
ok @mbjoseph i've made changes as requested in this PR. all tests are passing now!! please have another look.

mbjoseph:
Looks great @lwasser! :100:

lwasser:
woo hoo!! thank you so much @mbjoseph !!!

mirob9363:
It is working now! Hooray!

Issue #284, opened 2019-03-22
read the docs build for earthpy is failing quietly


lwasser:
Earthpy currently does not have a RTD yml file that will create a custom envt to support building earthpy docs. This file was added but then hte build began to fail quietly - the envt wouldnt complete install.
It seems like geopandas discovered a similar issue - see: 

* https://github.com/geopandas/geopandas/pull/932
* https://github.com/rtfd/readthedocs.org/issues/5220

A few things to try include

1. pinning the versions of each package to make the install easier and quicker
2. creating the envt file in the docs folder so it's specific to docs rather than to the broader earthpy dev envt. this allows us to be more specific. 
3. updated to the rtd yml file version 2.

lwasser:
closing this as the build seems to be working now given the updates.

PR #285, opened 2019-03-22
Fix Read the docs Build issue with an explicit envt file and v2 rtd.yml file

lwasser:
Please see: https://github.com/earthlab/earthpy/issues/284

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 - python=3.7
 - numpy
 - matplotlib=3.0.3
-- geopandas
+- geopandas=0.4.1
 - rasterio=1.0.21
 - scikit-image
 - shapely=1.6.4
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -14,3 +14,4 @@ dependencies:
 - rtree=0.8.3
 - sphinx=1.8.4
 - sphinx_rtd_theme=0.4.3
+- m2r=0.2.1
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,0 +1,16 @@
+name: earthpy_docs
+channels:
+- conda-forge
+dependencies:
+- python=3.7
+- numpy
+- matplotlib=3.0.3
+- geopandas
+- rasterio=1.0.21
+- scikit-image
+- shapely=1.6.4
+- fiona=1.8.4
+- pyproj=1.9.6
+- rtree=0.8.3
+- sphinx=1.8.4
+- sphinx_rtd_theme=0.4.3
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,0 +1,5 @@
+python:
+  version: 3
+  pip_install: true
+conda:
+  file: environment.yml
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -5,6 +5,6 @@ conda:
 python:
    version: 3.7
    install:
-      - requirements: requirements.txt
+      - requirements: dev-requirements.txt
       - method: pip
         path: .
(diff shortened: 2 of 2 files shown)


lwasser:
@mbjoseph there are some issues with RTD that others have experienced as well. In this PR, i

1. create an envt file specifically for RTD. it specifies EACH VERSION of each package that we need to build the docs. This makes RTD happy
2. it also bumps up the confit file to v2 vs v1. 

I'm noticing matplotcheck is also failing - presumably for the same reasons. 

If this looks ok to you, please review / approve it and i'll merge it accordingly :)

lwasser:
IMPORTANT: before this is merged, we need to switch RTD back to building off of master :) it's building off of this branch now just as a test.

mbjoseph:
Thanks for digging into this @lwasser! Having to maintain our requirements in yet another file seems like it could be a burden in the long run. Do you know:

- what specifically was not working with our current environment yaml file? (any links to relevant RTD log files would be helpful)
- whether we can edit the current yaml file to fix the issue?

I'm happy to experiment with this more too if that makes sense.

mbjoseph:
Also in the future, it's possible to turn on RTD for a branch without changing the "latest" version. If you go here: https://readthedocs.org/projects/earthpy/versions/ and under "Inactive Versions" find the branch you're working on, then click edit and click the "Active:" checkbox.

lwasser:
@mbjoseph when RTD tries to install the envt, it takes a long time trying to find each package. when this happens the envt install hangs and fails. this is the fix that was implemented by a few folks including geopandas (see links in the issue #284 ) essentially what i read is that it's faster for RTD and conda to find a specific version over having to dig for the version. matplotcheck is failing now for the same reason. if you find a work around please feel free to add to this PR. 

we can def edit the current yaml file to address this issue. but you will also notice i added some versions for a few sphinx things in the yaml file ... maybe we dont' need that. i didnt have time to test it but it wasn't finding m2r so i added it for example.

 i wanted the docs to be working live which is why i  switched them to this branch! i didn't want to merge into master without a discussion however.

mbjoseph:
Ok thanks @lwasser! I didn't realize that's why you were pinning package versions. A couple of questions:

For this PR: 

- How would you feel about setting `fail_on_warning: True` in the RTD config file? The current docs build is "succeeding" despite warnings that our modules can't be imported. It would be better if we got alerted of build failures, and maybe this is a way to do that: https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

Longer term, I have some questions about how we can mitigate the long term tech debt from adding another environment file  (e.g., how we update it, and how we test against it locally), but we can sort that out in the issues.

lwasser:
hey @mbjoseph sure let's add that fail on warning and see what happens? it would be good to get a notification when things fail if that is possible. 

i agree on the envt issues. it could be that if i remove the dev-requirements.txt file from the gui and keep all of the setup in the yaml file that it would properly find the sphinx extensions needed. it just is very slow to rebuild when you use the yaml file so i ran out of time friday. feel free to 

1. add the `fail_on_warning` flag to this pr and 
2. please around with the envt. perhaps another issue is a good idea so we can get working docs for now but know that some cleanup is needed. 

most other packages don't have 2 environments btw. i just did it because i didn't want to mess with our main dev envt given all of the issues we've had.

mbjoseph:
Ok @lwasser I added the fail on warning flag to the yaml file, and the build succeeds. https://readthedocs.org/projects/earthpy/builds/8810362/

I think we can merge this now. Once earthpy is on conda-forge, we won't need the other environment file, so the dual environment file issue should go away at some point.

lwasser:
sounds good @mbjoseph ... but let's see... just so you know i had a hard time getting docs to build with matplotcheck without the yml file!! but let's see what happnes. i'll go ahead and merge and will update RTD config now too!! thank ou so much!!

Issue #282, opened 2019-03-21
Bug with ep.plot_rgb(stretch=True)


nkorinek:
**Describe the bug**
When plotting landsat RGB bands with `plot_rgb()`, I tried to apply a stretch, since the image was "washed out". When I did `ep.plot_rgb(array, stretch=True)`, I get a warning saying "RuntimeWarning: overflow encountered in short_scalars: crange = cmax - cmin". 

Addionally, `es.bytescale()`, which is used in `ep.plot_rgb()`, gives the following error that stops the plot: "ValueError: `cmax` should be larger than `cmin`." 

When looking into this, I did `array.max()` and `array.min()`, which are the default cmax/cmin values. The maximum was 13016, and the minimum was -9999, which is a N/A value. There were other negative values in the array besides -9999, I'm not sure what they represented. 

To get around this bug, before running the `ep.plot_rgb()` function, I ran `output = es.bytescale(array, cmin = 0)`. After this is run, if you run `ep.plot_rgb(output, stretch=True)` the expected behavior happens, with no errors. If cmin is set not to negative, but to the next lowest value in the input array, such as -1737 in this case, the array does not stretch well, but there are no errors. 

**To Reproduce**
This code block should reproduce the behavior:

```
import os
from glob import glob
import earthpy as et
import earthpy.spatial as es
import earthpy.plot as ep

data = et.data.get_data("cs-test-landsat")
os.chdir(os.path.join(et.io.HOME, 'earth-analytics'))

landsat_bands = "data/cs-test-landsat/LC08_L1TP_034032_20160621_20170221_01_T1_sr_band*[2-4]*.tif"
stack_bands = glob(landsat_bands)
stack_bands.sort()
array, raster = es.stack(stack_bands)
ep.plot_rgb(array, stretch=True)
```
To make the error go away, add `array = es.bytescale(array, cmin=0)` above the `ep.plot_rgb()` line. 

**Expected behavior**
I expected the stretch to be able to run without doing any modification of the bytescale beforehand. 

**Screenshots**
The error message when no modification is made to the bytescale:
![image](https://user-images.githubusercontent.com/38253680/54788290-1cdf2400-4bf4-11e9-914c-392acb77c2a6.png)

The image when the cmin is set to the lowest value in the array that isn't an N/A value (-1737, in this case) before running the rgb plot.
![image](https://user-images.githubusercontent.com/38253680/54788523-e6ee6f80-4bf4-11e9-8d07-76e6e1a2b892.png)

[...]

PR #287, opened 2019-03-25
fix bug in plot rgb when user forgets to deal with no data values + bytescale fix + test cleanup

lwasser:
Please see https://github.com/earthlab/earthpy/issues/282 

When a user provides and RGB image to plot_rgb but leaves the no data values in, the result is an error with bytescale if the no data val is small (e.g. -9999). Here, we check to ensure the max-min is not less than 0 BEFORE bytescale fails on them. We will implement this in conjunction with an upgrade to `stack_bands` to allow for no data values in hopes users won't forget to handle na values when plotting. 

UPDATE: through this pr i will

1. fix a bug in bytescale to ensure range is calculating properly via floating point vals.
2. break up bytescale tests into a new file AND also individual tests - they are currently all in one ginormous function
3. add some notes to the docs for plot_rgb to ensure users clean our no data values.


lwasser:
weird things
1. this test
```
def test_1band(rgb_image):
    """Test to ensure the input image has 3 bands to support rgb plotting.
    If fewer than 3 bands are provided, fail gracefully."""
    a_rgb_image, _ = rgb_image

    with pytest.raises(ValueError, match="Input needs to be 3 dimensions"):
        plot_rgb(a_rgb_image[1])
    plt.close()
``` 
tests for a value error yet [code cov](https://codecov.io/gh/earthlab/earthpy/pull/287/src/earthpy/plot.py?before=earthpy/plot.py#L299) is noticing that the message part of the error is NOT tested. not sure why. the function name is unique.

@mbjoseph in theory this pr is good to go. code cov is not happy and i'm not sure why. the above is one line that it thinks is not being hit but is being hit (i think?). anyway please have a look and let me know what you think!!

earthpy/plot.py
```diff
@@ -327,7 +328,8 @@ def plot_rgb(
     ax.imshow(rgb_bands, extent=extent)
     ax.set_title(title)
     ax.set(xticks=[], yticks=[])
-    return fig, ax
```
mbjoseph: This seems like a substantial change in behavior - is this related to #282? 

If users are trying to use this function to make multipanel plots or use the `fig` object, will this change prevent them from doing so?

earthpy/tests/test_bytescale.py
```diff
@@ -0,0 +1,89 @@
...
+import earthpy.spatial as es
+
+
+@pytest.fixture
+def byte_arr():
+    return np.random.randint(300, size=(10, 10))
+
+
+def test_high_val_range():
+    """A 16 bit int arr with values at the end of the range, should be scaled properly.
```
mbjoseph: Each line in this docstring goes over the 79 character limit

earthpy/tests/test_bytescale.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Note that this test explicitly hits the user case of someone providing an array of a
+    dtype at the end of it's range of values. Bytescale should not fail in this case.
+    """
+
```
mbjoseph: Good place for a comment to explain where these magic numbers come from
```suggestion
    # The range of int16 values in python is -32768 to 32767
```

earthpy/tests/test_bytescale.py
```diff
@@ -0,0 +1,89 @@
...
+    # High value is less than low value
+    with pytest.raises(
+        ValueError, match="`high` should be greater than or equal to `low`."
+    ):
+        es.bytescale(byte_arr, high=100, low=150)
+
+
+def test_low_high_vals_work(byte_arr):
+    """A proper high and low value argument should return an arr with thos vals as the min and max."""
```
mbjoseph: This seems like it could be condensed

```suggestion
    """ The high and low args set the output min and max. """
```

earthpy/tests/test_bytescale.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        es.bytescale(byte_arr, cmin=100, cmax=50)
+
+
+def test_cmax_equals_cmin(byte_arr):
+    """Fail gracefully when the cmax is smaller than the cmin."""
+
+    with pytest.raises(
+        ValueError,
+        match="`cmax` and `cmin` should not be the same value. Please specify `cmax` > `cmin`",
```
mbjoseph: This line is longer than it needs to be - `match` uses regex so we can do:

```suggestion
        match="`cmax` and `cmin` should not be the same value",
```

earthpy/tests/test_bytescale.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        es.bytescale(byte_arr, cmin=100, cmax=100)
+
+
+def test_cmax_cmin_work(byte_arr):
+    """"Cmax and min values should yield and arr with the same min and max values."""
```
mbjoseph: Can also be condensed and clarified:

```suggestion
    """" cmin and cmax should not change the output min and max. """
```

earthpy/tests/test_plot_rgb.py
```diff
@@ -26,14 +26,25 @@ def rgb_image():
     return rgb, ext
 
 
+def test_no_data_val(rgb_image):
+    """An array with a large no data value should not print nicely, but it should work."""
```
mbjoseph: This docstring doesn't really describe what's being tested (e.g., we aren't testing that the array is not printing nicely).

Can it be clarified and shortened to fit with our line length limits? e.g., 

```suggestion
    """ Arrays containing -9999 make plots when stretched. """
```

earthpy/plot.py
lwasser: Nope. 
1. you can always access the figure object using ax.figure if need be.
BUT if a user wants to implement multi-panel, they need to setup the plot themselves and then use the `ax=` parameter ... 

this test
```
def test_ax_provided(rgb_image):
```
tests that a user can in fact provide a custom axis and things still work!

mbjoseph: Sweet!

lwasser:
ok @mbjoseph travis looks happy - the github web hook doesn't seem to be updating but when i go to the travis build it is passing. I think i've covered all of your feedback here but let me know. I wasn't sure what you meant about `plt.show()` as i am calling it in the function... but if you help me better understand what you are looking for i can likely account for it!!

mbjoseph:
Looks good @lwasser!

lwasser:
yahoo! thanks @mbjoseph !!!

Issue #219, opened 2019-02-07
Refactoring the tests - one test file per function (or function + helpers) 


mbjoseph:
We started out defining one test module per earthpy module, and then transitioned to defining one test module per function or group of functions in earthpy modules. We should choose one approach. 

To be consistent, we could refactor the tests so that one function or groups of related functions are tested in one test module, rather than using  a one to one mapping of earthpy modules to test modules. @lwasser is that your preference?

lwasser:
We will reorganize the tests so there is one file named `test_func_name.py` per function!!

lwasser:
this is done!!

Issue #218, opened 2019-02-07
Purge module-specific test fixtures from tests/conftest.py


mbjoseph:
Any fixtures defined in tests/conftest.py are available for all tests. This is great when needed, but currently there are some fixtures in conftest.py that are only used and needed by one test module. These fixtures should be defined in that test module, instead of conftest.py. This has two benefits: 

1. Less clutter in conftest.py
2. Easier to reason about test fixtures that are defined and used in the same module

lwasser:
let's hold off on this until we implement -- https://github.com/earthlab/earthpy/issues/219

lwasser:
this has been implemented as well!! closing!!

PR #303, opened 2019-04-03
Test Cleanup & Re-org

lwasser:
This PR addresses https://github.com/earthlab/earthpy/issues/219  and #218 
it includes:
1. cleanup of `conftest.py` file to ensure only fixtures used across multiple functions are included
2. splitting up the `test_spatial.py` file into function specific test files.

In this cleanup process, i noticed that the extent_to_json() function has ALL TESTS in one giant function. So that is cleaned up as well.

### DIFF
--- a/earthpy/tests/test_extent_to_json.py
+++ b/earthpy/tests/test_extent_to_json.py
@@ -1,0 +1,39 @@
+""" Tests for the extent to json function """
+
+
+import pandas as pd
+import pytest
+import geopandas as gpd
+from shapely.geometry import Polygon, Point
+import earthpy.spatial as es
+
+
+def test_extent_to_json():
+    """"Unit tests for extent_to_json()."""
+    # Giving a list [minx, miny, maxx, maxy] makes a polygon
+    list_out = es.extent_to_json([0, 0, 1, 1])
+    assert list_out["type"] == "Polygon"
+
+    # The polygon is the unit square
+    list_poly = Polygon(list_out["coordinates"][0])
+    assert list_poly.area == 1
+    assert list_poly.length == 4
+
+    # Providing a GeoDataFrame creates identical output
+    points_df = pd.DataFrame({"lat": [0, 1], "lon": [0, 1]})
+    points_df["coords"] = list(zip(points_df.lon, points_df.lat))
+    points_df["coords"] = points_df["coords"].apply(Point)
+    gdf = gpd.GeoDataFrame(points_df, geometry="coords")
+    gdf_out = es.extent_to_json(gdf)
+    assert gdf_out == list_out
+
+    # Giving non-list or GeoDataFrame input raises a ValueError
+    with pytest.raises(ValueError):
+        es.extent_to_json({"a": "dict"})
+
+    # Giving minima that exceed maxima raises an error for both x and y coords
+    with pytest.raises(AssertionError):
+        es.extent_to_json([1, 0, 0, 1])
+
+    with pytest.raises(AssertionError):
+        es.extent_to_json([0, 1, 1, 0])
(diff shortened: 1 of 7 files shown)


lwasser:
ok @mbjoseph this PR addresses two open issues surrounding test cleanup in earthpy. Please have a look and let me know what you think!! Note that i didn't look closely at test content here. I just looked at the data management / cleanup components. one function however had a ginormous set of tests so i did break those out (`test_extent_to_json.py`)

earthpy/tests/test_extent_to_json.py
```diff
@@ -0,0 +1,56 @@
...
+import earthpy.spatial as es
+
+
+@pytest.fixture
+def list_out():
+
+    return es.extent_to_json([0, 0, 1, 1])
+
+
+def test_list_format_works():
```
mbjoseph: Can this test could use the `list_out` fixture you created?
```suggestion
def test_list_format_works(list_out):
```

earthpy/tests/test_extent_to_json.py
```diff
@@ -0,0 +1,56 @@
...
+    """" Giving a list [minx, miny, maxx, maxy] makes a polygon"""
+    list_out = es.extent_to_json([0, 0, 1, 1])
```
mbjoseph:
```suggestion

```

earthpy/tests/test_extent_to_json.py
```diff
@@ -0,0 +1,56 @@
...
+import pandas as pd
+import pytest
+import geopandas as gpd
+from shapely.geometry import Polygon, Point
+import earthpy.spatial as es
+
+
+@pytest.fixture
+def list_out():
+
```
mbjoseph: Maybe a comment here to explain and fill this blank line?
```suggestion
    """ A JSON extent on the unit square spanning (0, 0), (1, 1). """
```

earthpy/tests/test_extent_to_json.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_not_a_list():
+    """Providing a non-list or GeoDataFrame input raises a ValueError"""
+
+    with pytest.raises(ValueError):
+        es.extent_to_json({"a": "dict"})
+
+
+def test_min_exceeds_max():
```
mbjoseph: I didn't check the error message or know about parameterized tests when I first wrote these tests, but this is a good candidate, because it's almost the same test for two different inputs:

```python
@pytest.mark.parametrize(
    "list_input,error",
    [
        ([1, 0, 0, 1], "xmin must be <= xmax"),
        ([0, 1, 1, 0], "ymin must be <= ymax"),
    ],
)
def test_min_exceeds_max(list_input, error):
    """Giving minima that exceed maxima raises error for both x and y coords"""

    with pytest.raises(AssertionError, match=error):
        es.extent_to_json(list_input)

```

lwasser: cool @mbjoseph can you kindly help me understand what this means?? what is the `mark.parametrize` decorator or what does it do??

earthpy/tests/test_extent_to_json.py
lwasser: what do you think about the var name `list_out` @mbjoseph it just occurred to me that it's pretty darn uninformative / generic.

earthpy/tests/test_extent_to_json.py
mbjoseph: It provides multiple sets of inputs to the same test. So this test will be executed twice, once with the first set of inputs, and then again with the second: https://docs.pytest.org/en/latest/parametrize.html#parametrize-basics

lwasser:
Thank you @mbjoseph !! it's nice to get all of the tests cleaned up !!! I appreciate all of you improvements on them!!

Issue #302, opened 2019-04-03
Time to update the readme with better install instructions


lwasser:
The current readme

* still references the old pip install from github instructions
* doesn't mention conda-forge (yay!)
* is outdated.

Update the readme to present a friendly overview of how to install and use EarthPy

mbjoseph:
@lwasser here's a checklist I stole from https://github.com/noffle/art-of-readme

- [x] One-liner explaining the purpose (A package that makes it easier to plot, and manipulate spatial data in Python.)
- [ ] Necessary background context & links (e.g., we wrote this package because xyz...)
- [ ] Potentially unfamiliar terms link to informative sources
- [x] Installation instructions
- [ ] Clear, runnable example of usage (not sure what this should be, but it should be short and give a sense for what EarthPy does)
- [ ] Performs cognitive funneling (start broad at top, and have detail toward the bottom)
- [ ] Caveats and limitations 
- [ ] Doesn't rely on images to relay critical information
- [ ] License info

Some specific things that we might consider: 

- Link to contributing docs 
- Remove text about testing and making docs (this is now in the contributing docs)
- Move **Data generated for testing** to contributing docs (that content is not yet in the contributing docs)

lwasser:
i think that for the usable examples, we should build out vignettes and link to them directly. i've started this process. then each time we add a vignette we can update the readme!

lwasser:
ok i am closing this given the above merged PR. if there are other readme items that require updates, let's open a shiny new issue!!

PR #304, opened 2019-04-04
start at readme update

lwasser:
This begins the process of updating the readme file.
see:
https://github.com/earthlab/earthpy/issues/302

@mbjoseph what items should we include in the readme? it seems to me that there is some information in the contributing doc that was started here. maybe we can just include links to files as needed? thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,39 @@
 ![PyPI - Downloads](https://img.shields.io/pypi/dm/earthpy.svg?color=purple&label=pypi%20downloads&style=plastic)
 ![Conda](https://img.shields.io/conda/dn/conda-forge/earthpy.svg?color=purple&label=conda-forge%20downloads&style=popout)
 
-A package that makes it easier to plot, and manipulate spatial data in Python.
+EarthPy is makes it easier to plot and manipulate spatial data in Python.
+
+## Why EarthPy?
+
+Python is a generic programming language designed to support many different applications. Because of this, many commonly
+performed spatial tasks for science including plotting and working with spatial data take many steps of code. EarthPy 
+takes advantage of functionality developed for raster data (rasterio) and vector data (geopandas) and simplifies the 
+code needed to :
+
+* Stack raster bands from data such as Landsat into an easy to use numpy array
+* Plot rgb and other 3 band combination images
+* View histograms of sets of raster 
+* Create discrete (categorical) legends
+
+EarthPy also has an io module that allows users to 
+
+1. Quickly access pre-created datasubsets used in the earth-analytics courses hosted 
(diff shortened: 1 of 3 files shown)


mbjoseph:
@lwasser I added some suggestions in #302 so we can keep the discussion in the issues (plus we might need to spread the overhaul over multiple PRs).

lwasser:
@all-contributors please add @mbjoseph for being cool

lwasser:
@all-contributors please add @mbjoseph for infrastructure coolness

lwasser:
ok giving up on the pipe dream for the time being given the contributors format is HTML right now and i'm not sure that RTD can handle HTML within md and html files. if it can we can use includes however but i just don't know how to do it.

lwasser:
ok @mbjoseph i've played around with a few things. here's the deal

1. the contributor bot is cool BUT... the issue is that it creates a contrib table in HTML format. and that clashes with RTD unfortunately unless there is some way to include html. this is just fine for our readme as markdown can tolerate HTML (our website uses both) but .rst doesn't like it.
2. so let's look at the readme text and such only in this PR. and maybe i'll keep exploring other ways to update the contributors. 

There could be a clever way to create a little script to do it via the GH api. it wouldn't be hard. but i'd rather find something we can just use!!

README.md
```diff
@@ -7,64 +7,104 @@
...
+EarthPy also has an io module that allows users to 
+
+1. Quickly access pre-created datasubsets used in the earth-analytics courses hosted 
+on [www.earthdatascience.org](https://www.earthdatascience.org) 
+2. Download other datasets that they may want to use in their workflows.
+
+## View Example EarthPy Applications in Our Documentation Gallery  
+
+Check out our [vignette gallery](https://earthpy.readthedocs.io/en/latest/gallery_vignettes/index.html) for 
+applied examples of using earthpy in common spatial workflows.
```
mbjoseph:
```suggestion
applied examples of using EarthPy in common spatial workflows. 
```

README.md
```diff
@@ -7,64 +7,104 @@
...
-yet you can still use the `--upgrade` argument.
+To install, use `pip` or `conda-forge`. We encourage you to use `conda-forge` if you are a conda users. 
 
-```bash
-$ pip install --upgrade git+https://github.com/earthlab/earthpy.git
-```
+### Install via Pip
 
-Then import it into python.
+To install earthPy via `pip` use:
```
mbjoseph:
```suggestion
To install EarthPy via `pip` use:
```

lwasser:
ok @mbjoseph i just merged all of your suggested commits!!

PR #312, opened 2019-04-11
Another bug fix in plot rgb

lwasser:
well it turns out that i missed yet another bug in plot_rgb.

This fix addresses the issue that when you attempt to create a multi-panel plot, this function was plotting the FIRST plot and the rest of the panels were empty. This was because we decided to add `plt.show()` to the function call.

@mbjoseph  when you look at this PR you'll notice there is now some redundant code in the last statement where the plot is generated. 

If we don't want that code, we need to 

1. drop the `plt.show()` call altogether. 

i could be convinced to do this given the function returns the ax object which "prints" as a returned item. or we leave it nice and clean and it just plots for the user by default UNLESS they are creating a more involved plot. 

Note that in this i'm adding a second test that tests a multi panel setup . i was testing only ONE axis previously which was passing just fine but failing upon addding the second axis (which was empty!!).

### DIFF
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -323,12 +323,17 @@ def plot_rgb(
     # Then plot. Define ax if it's default to none
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
+        ax.imshow(rgb_bands, extent=extent)
+        ax.set_title(title)
+        ax.set(xticks=[], yticks=[])
+        # Multipanel won't work if plt.show is called prior to second plot def
+        # Thus we either have redundant code or DON"T call plt.show
+        plt.show()
     else:
-        fig = None
-    ax.imshow(rgb_bands, extent=extent)
-    ax.set_title(title)
-    ax.set(xticks=[], yticks=[])
-    plt.show()
+        # fig = None
+        ax.imshow(rgb_bands, extent=extent)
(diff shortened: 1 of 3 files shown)


earthpy/plot.py
```diff
@@ -323,12 +323,17 @@ def plot_rgb(
     # Then plot. Define ax if it's default to none
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
+        ax.imshow(rgb_bands, extent=extent)
+        ax.set_title(title)
+        ax.set(xticks=[], yticks=[])
+        # Multipanel won't work if plt.show is called prior to second plot def
+        # Thus we either have redundant code or DON"T call plt.show
+        plt.show()
```
lwasser: @mbjoseph this is the problem area. when you call plt.show() too early it populates all of the panels they end up with no data in them. so this is a workaround. 

conversely we remove plt.show() and just call this part

```
ax.imshow(rgb_bands, extent=extent)
        ax.set_title(title)
        ax.set(xticks=[], yticks=[])
```
once

lwasser: UPDATE: i figured out a cleaner way to implement this!!

lwasser:
thank you @mbjoseph  just updated the changelog and now am merging!!!

Issue #314, opened 2019-04-11
plot_bands vs single image plot


lwasser:
Well -- currently we have a function called plot_bands that will plot several bands in a stacked numpy array. this worked well but recently i've been using it for single band plots too. However there is a limitation as currently designed where when plotting one band, you may want to overlay another layer (say a fire boundary or something else) on top of that band. Or you might also want to plot in your own multi-panel configuration. 

Currently this is not supported when plotting a single band. The question is - how should we implement this. One option is to create a simple function that plots ONE BAND with all of the options that we want. Then, the `plot_bands()` function could call that in a loop.

The bigger question is does plot_bands support single and multi-band plotting OR do we create two separate functions (even if plot_bands calls the second function). 

Regardless i think a function could be good to reduce repetition in the function -- specifically this is repeated twice: 

```
           if scale:
                arr_im = es.bytescale(arr_im)

            im = ax.imshow(
                arr_im, cmap=cmap, vmin=vmin, vmax=vmax, extent=extent
            )
            if title:
                ax.set(title=title[i])
            else:
                ax.set(title="Band %i" % band)
            if cbar:
                colorbar(im)
            ax.set(xticks=[], yticks=[])
```

lwasser:
Here is an example of what plotting looks like now
```
# force the z axis min an dmax to be the same
fig, (ax1, ax2) = plt.subplots(2, 1, figsize=(12, 10))
ndvi_2015 = ax1.imshow(naip_ndvi_2015,
                       cmap='PiYG',
                       vmin=-1, vmax=1,
                       extent=naip_2015_ext)
fire_bound_utmz13.plot(ax=ax1, color='None',
                       edgecolor='black', linewidth=2)
ep.colorbar(ndvi_2015)
ax1.set(title="NAIP Derived NDVI\n September 2015 pre-fire\n This is NOT a homework plot but here for you to compare your answer if you want.")

ndvi_2017 = ax2.imshow(naip_ndvi_2017,
                       cmap='PiYG',
                       vmin=-1, vmax=1,
                       extent=naip_2015_ext)
fire_bound_utmz13.plot(ax=ax2, color='None', edgecolor='black', linewidth=2)
ep.colorbar(ndvi_2017)
ax2.set(title="NAIP Derived NDVI\n September 2017 post-fire")
# ax2.text(457163.0, 4424640.0,"hello")
plt.tight_layout()
ax1.set_axis_off()
ax2.set_axis_off()
plt.show()
```

[...]

PR #316, opened 2019-04-13
Adding multi panel support to plot_bands

lwasser:
addresses https://github.com/earthlab/earthpy/issues/314

### DIFF
--- a/earthpy/plot.py
+++ b/earthpy/plot.py
@@ -202,19 +202,6 @@ def plot_bands(
             band = i + 1
 
             arr_im = arr[i]
-            # if scale:
-            #     arr_im = es.bytescale(arr_im)
-
-            # im = ax.imshow(
-            #     arr_im, cmap=cmap, vmin=vmin, vmax=vmax, extent=extent
-            # )
-            # if title:
-            #     ax.set(title=title[i])
-            # else:
-            #     ax.set(title="Band %i" % band)
-            # if cbar:
-            #     colorbar(im)
-            # ax.set(xticks=[], yticks=[])
 
             if title:
                 the_title = title[i]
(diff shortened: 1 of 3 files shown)


lwasser:
@mbjoseph  this puppy is finally ready for review. cleaned up and all tests passing!!

earthpy/plot.py
```diff
@@ -112,11 +168,8 @@ def plot_bands(
 
     Returns
     ----------
-    tuple
-        fig : figure object
-            The figure of the plotted band(s).
-        ax or axs : axes object(s)
-            The axes object(s) associated with the plot.
+    ax or axs : axes object(s)
+        The axes object(s) associated with the plot.
```
mbjoseph: Does it make sense to add some specificity here by saying what class the returned axes are? e.g., does this return an array of matplotlib.axes._subplots.AxesSubplot objects, or something else?

earthpy/plot.py
```diff
@@ -71,6 +71,61 @@ def colorbar(mapobj, size="3%", pad=0.09):
     return fig.colorbar(mapobj, cax=cax)
 
 
+def _plot_image(
```
mbjoseph: Nice work refactoring this out into its own function!

lwasser: thank you @mbjoseph !! i had to think about it for a while and then finally realized how to implement it more efficiently! it actually makes managing the entire main function easier i just didn't know how to do it at first as my brain was making it more complicated than it needed to be!!

earthpy/plot.py
lwasser: totally!

lwasser:
ok @mbjoseph i've made that change to the docstring to clarify!! please let me know if you catch anything else with this PR!

lwasser:
@mbjoseph this is failing one of the appveyor tests... it's because of a download that is not related to this PR and i suspect rebuilding the PR would fix it (i've been doing that with travis sometimes) but i don't see how i can reset a build in appveyor. thoughts on what to do here? review / merge or try to rebuild it i just don't know how.

PR #317, opened 2019-04-22
Clip multi point line and polygon features 

lwasser:
This PR adds functionality for earthpy to support clipping multi part features.


earthpy/clip.py
```diff
@@ -6,6 +6,9 @@
 
 """
 
+import pandas as pd
+import geopandas as gpd
+
 # TODO: Clip poly should use OVERLAY not spatial indexing + intersects
```
lwasser: This line can be removed now!

earthpy/clip.py
```diff
@@ -74,6 +77,45 @@ def _clip_line_poly(shp, clip_obj):
...
+    -------
+    GeoDataFrame
+        The returned GeoDataFrame is a clipped subset of shp
+        that intersects with clip_obj.
+    """
+
+    # Clip multi polygons
+    clipped = _clip_line_poly(shp.explode().reset_index(level=[1]), clip_obj)
+    # If there are lines and poly's you can't just dissolve
+    if any(clipped.geometry.type == "MultiLineString"):
```
lwasser: This will fail if there are no multilines in the output.. refactor accordingly.

lwasser:
ok @joemcglinchy can you please test the NEW clip fnction that supports multi polygons for us?? this PR is the one to test.

mbjoseph:
@nkorinek @lwasser thanks for opening the PR. I haven't taken a deep dive yet, but I am noticing two issues that would be good to resolve right off the bat: 

1. There is a decrease in code coverage associated with this PR. Please fix this by looking at which lines are not covered on codecov (e.g., https://codecov.io/gh/earthlab/earthpy/compare/4d65d5e67f886bc7c5e48cfdba8abc5e3e00462d...e072469f8d5c5de05312b64969e440760e62e411). I suspect this is due in part to duplication in the test modules (see `multi_gdf` for example). 

2. There are TODO items that are added as code comments as part of this PR. It's hard to tell which of these are up to date, and it will be hard long term to track the progress of these issues if we try to keep all relevant info in the source code. For these reasons, it would be good to transition these TODO comments to issues, so that we can track the things we need to do separately from the actual source code.

lwasser:
hey @mbjoseph i'll have a look. i had thought i had cleaned up the dup fixtures but it sounds like i missed some. the code cov thing is interesting because clip is at 100% but let's see if the dup removal fixes things.

lwasser:
ok @mbjoseph i think this should be much cleaner now. my apologies for the mess. i thought i had cleaned things up but didn't realize i could check for dups in codecov. that's really good to know for future. TODO's and dups are not cleaned up and all tests are passing!

nkorinek:
@lwasser Thanks for making all of those changes, they look really good! What still need to be done for this pull request on my end? I'm trying to go through everything you did to see how you fixed my issues, but is there any more actual code that I need to modify?

lwasser:
@nkorinek thank you for all of the work on this!! i think you can focus on the vignettes for the time being. when we meet tomorrow i'll have some other things for you to work on related to other vignettes and the lessons!!

earthpy/clip.py
```diff
@@ -34,6 +35,41 @@ def _clip_points(shp, clip_obj):
...
+        Reference polygon for clipping.
+
+    Returns
+    -------
+    GeoDataFrame
+        The returned GeoDataFrame is a clipped subset of shp
+        containing multi-point and point features.
+    """
+
+    if (
```
mbjoseph: This has some duplication (`shp["geometry"].iloc[0].type`) but you could do something like the following: 

```python
if "Point" in shp["geometry"].iloc[0].type
```

earthpy/clip.py
```diff
@@ -34,6 +35,41 @@ def _clip_points(shp, clip_obj):
...
+        shp["geometry"].iloc[0].type == "Point"
+        or shp["geometry"].iloc[0].type == "MultiPoint"
+    ):
+        # This line works
```
mbjoseph: Delete this comment?

earthpy/clip.py
```diff
@@ -74,6 +110,65 @@ def _clip_line_poly(shp, clip_obj):
     return clipped[clipped.geometry.notnull()]
 
 
+def _clip_multi_poly_line(shp, clip_obj):
```
mbjoseph: I agree the implementation seems a wee bit hacky. Can it be simplified by relying on the concatenation of zero row data frames instead of conditional logic? e.g., 

```python
def _clip_multi_poly_line(shp, clip_obj):
    """..."""
    # Clip multi polygons
    clipped = _clip_line_poly(shp.explode().reset_index(level=[1]), clip_obj)

    lines = clipped[
        (clipped.geometry.type == "MultiLineString")
        | (clipped.geometry.type == "LineString")
    ]
    line_diss = lines.dissolve(by=[lines.index]).drop(columns="level_1")

    polys = clipped[clipped.geometry.type == "Polygon"]
    poly_diss = polys.dissolve(by=[polys.index]).drop(columns="level_1")

    return gpd.GeoDataFrame(
        pd.concat([poly_diss, line_diss], ignore_index=True)
    )
```

Note that I tested this locally and all tests pass. If this simpler function is incorrect in some way, we should add some tests to capture that.

earthpy/clip.py
```diff
@@ -146,14 +241,14 @@ def clip_shp(shp, clip_obj):
...
-        raise ValueError(
-            "Clip doesn't currently support multipart geometries. Consider "
-            "using .explode to create unique features in your GeoDataFrame"
-        )
-
-    if shp["geometry"].iloc[0].type == "Point":
+    # if there are multi points
+    if any(shp.geometry.type == "MultiPoint"):
+        return _clip_multi_point(shp, clip_obj)
+    elif shp["geometry"].iloc[0].type == "Point":
```
mbjoseph: I'm noticing some places where we use `shp["geometry"].iloc[0].type` instead of `shp.geometry.type`. The former seems somewhat brittle because we only look at the first row. Is there a good reason for us to do this, or should we change this to 

```python
elif any(shp.geometry.type == "Point"):
```

earthpy/tests/test_clip.py
```diff
@@ -19,33 +19,47 @@ def make_locs_gdf():
...
+    )
+    gdf["attr2"] = ["study area"]
     return gdf
 
 
 def make_locs_buff():
     """ Buffer points to create multi poly. """
     buffered_locations = make_locs_gdf()
     buffered_locations["geometry"] = buffered_locations.buffer(4)
+    buffered_locations["type"] = ["plot", "plot", "plot", "plot"]
```
mbjoseph:
```suggestion
    buffered_locations["type"] = "plot"
```

That'll do the same thing more compactly

earthpy/tests/test_clip.py
```diff
@@ -19,33 +19,47 @@ def make_locs_gdf():
...
+        make_single_rect_poly_gdf(),
+        how="symmetric_difference",
     )
     return donut
 
 
 @pytest.fixture
 def locs_gdf():
-    """ Get a dummy point GeoDataFrame.
+    """ Create a sample point GeoDataFrame.
```
mbjoseph: Nice :pray:

earthpy/tests/test_plot_bands.py
```diff
@@ -17,6 +17,11 @@ def one_band_3dims():
     )
 
 
+@pytest.fixture
+def one_band_3dims():
```
mbjoseph: How did this get in here?

lwasser: @mbjoseph i was cleaning up duplicate fixtures and noticed issues with conftest. i then realized there were still fixtures in conftest that should be in a specific test module. so i moved it. do you want me to move this over to another pr? it's such a minor change and it was part of my larger clean up all tests effort in this work ... so that is what happened!!

i'll work on the other edits over the next few days!! thank you so much for these.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [unreleased]
```
lwasser: @nkorinek this is an example of updating the changelog. this is what i'd like you to do for each vignette that you submit.

earthpy/tests/test_plot_bands.py
mbjoseph: @lwasser I think this is a duplicate fixture that has yet to be cleaned up. It already exists on the master branch in this file, and on this branch the fixture `one_band_3dims` is defined twice now in `test_plot_bands.py`.

earthpy/clip.py
lwasser: ok.. the above didn't work but i could do it using a list comprehension. i'm slightly worried about how that will scale but we know that this doesn't scale as well as it could in parallel anyway and that people are working on that for geopandas so for now a list comprehension will do!

earthpy/clip.py
lwasser: done!!

earthpy/clip.py
mbjoseph: That's weird - I implemented that change and all of the tests still pass, i.e., 

```python
def _clip_multi_point(shp, clip_obj):
    """..."""

    if "Point" in shp["geometry"].iloc[0].type:
        clipped = _clip_points(shp.explode().reset_index(level=[1]), clip_obj)
        clipped = clipped.dissolve(by=[clipped.index]).drop(columns="level_1")[
            shp.columns.tolist()
        ]
    return clipped
```

But, on second thought, would this function make more sense without the if statement? If the condition is not satisfied, then it will raise an exception when trying to return `clipped`, right?

lwasser: yes... well i realized two things that don't work
1. it just looks at the first geom_type but there could be many features so doing [0] skips that . thus a list comprehension is desired.
BUT
2. i just realized when the function is called, it is already assuming there is atleast one multipoint in the GDF SO... i'm thinking rather than the list comprehension should jsut drop the conditional altogether! 

```
    if any(shp.geometry.type == "MultiPoint"):
        return _clip_multi_point(shp, clip_obj)
```
so i thijnk it might make sense to just remove the check altogether for this particular helper function since it's being called on the condition that a multipoint feature exists already.

lwasser: Ok this has been cleaned up. i think the simpler version is better!! it won't hit this helper unless the conditional in the main function has been satisfy so there is no need to test it again as far as i can see.

earthpy/clip.py
lwasser: i'll test this. we have the edge case in our tests where a sliver (a line) and a poly are returned. Let me see if it works when we just have a poly.  although i did think i made a poly only specific test...

lwasser: gosh... so i didn't realize that zero-row dataframe concatenation would work @mbjoseph but it does!! thank you. this code is so much cleaner. i've implemented this. i have a few other questions for you WRT to fixtures that i'll add in a moment

earthpy/tests/test_clip.py
lwasser: it will. i again didn't realize i could do it that way and it would propagate to the entire DF!! thanks

earthpy/clip.py
lwasser: hmmm i suspect this is just a product of us working and reworking this for so long. it should be `any()` as that is much more robust!!

lwasser: actually this is an interesting potential glitch. if for some reason someone has a mix of shapes this could fail. this is very atypical as a shapefile can't support that BUT geopandas does support this. maybe this is a see what happens situation because it's an edge case but i could see it happening if someone does some funky processing first. i'll switch to any for the time being and then wait for any edge cases to come in as bugs IF they come in. in theory it shouldn't be an issue if people are using conventional shapefiles. BUT not sure about geojson support.

earthpy/tests/test_plot_bands.py
lwasser: oh you're right. this is my old version of this fixture. i'm not sure how all of these fixtures got duplicated... actually i have a suspicion ...

earthpy/tests/test_clip.py
```diff
@@ -19,33 +19,47 @@ def make_locs_gdf():
...
 
     This fixture calls make_locs_gdf(), which is a function that is used in
```
lwasser: @mbjoseph  this is where i have the question. i'm really not sure how to translate this in terms of cleaning up the fixtures here... can you somehow clarify for me? that discussion actually confused me more than it helped.

earthpy/tests/test_clip.py
```diff
@@ -19,33 +19,47 @@ def make_locs_gdf():
...
     buffered_locations["geometry"] = buffered_locations.buffer(4)
+    buffered_locations["type"] = "plot"
     return buffered_locations
 
 
 def make_donut_geom():
     """ Make a donut geometry. """
     donut = gpd.overlay(
-        make_locs_buff(), make_poly_in_gdf(), how="symmetric_difference"
+        make_locs_buff(),
```
lwasser: ok another question. @mbjoseph because i use these functions, it's actually harder to test the code because i have functions wrapped in functions wrapped in fixtures. just curious what your thoughts are on this? it's kind of tedious to maintain and troubleshoot.

lwasser:
ok @mbjoseph this is ready for another review but i also had some fixture questions for you. thanks!!

earthpy/tests/test_clip.py
mbjoseph: Sorry that discussion was not helpful. If you're looking to get rid of the helper functions, you could define some fixtures in terms of others, which would definitely simplify things, e.g., 

```python
"""Tests for the clip module."""

import pytest
import numpy as np
from shapely.geometry import Polygon, Point, LineString
import shapely
import geopandas as gpd
import earthpy.clip as cl


@pytest.fixture
def locs_gdf():
    """ Create a sample point GeoDataFrame.

    This fixture calls make_locs_gdf(), which is a function that is used in
    multiple fixtures. But, fixtures are not supposed to be used like that:

    see https://github.com/pytest-dev/pytest/issues/3950 for discussion
    """
    pts = np.array([[2, 2], [3, 4], [9, 8], [-12, -15]])
    gdf = gpd.GeoDataFrame(
        [Point(xy) for xy in pts],
        columns=["geometry"],
        crs={"init": "epsg:4326"},
    )
    return gdf


@pytest.fixture
def locs_buff(locs_gdf):
    """ Buffer points to create multi poly. """
    buffered_locations = locs_gdf
    buffered_locations["geometry"] = buffered_locations.buffer(4)
    buffered_locations["type"] = "plot"
    return buffered_locations


@pytest.fixture
def linez_gdf():
    """ Create Line Objects For Testing """
    linea = LineString([(1, 1), (2, 2), (3, 2), (5, 3)])
    lineb = LineString([(3, 4), (5, 7), (12, 2), (10, 5), (9, 7.5)])
    gdf = gpd.GeoDataFrame(
        [1, 2], geometry=[linea, lineb], crs={"init": "epsg:4326"}
    )
    return gdf


@pytest.fixture
def single_rect_poly_gdf():
    """ Fixture for a bounding box polygon. """
    poly_inters = Polygon([(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)])
    gdf = gpd.GeoDataFrame(
        [1], geometry=[poly_inters], crs={"init": "epsg:4326"}
    )
    gdf["attr2"] = ["site-boundary"]
    return gdf


@pytest.fixture
def smaller_clip_rect_poly_gdf():
    """ Bounding box from (-5, 5) to (15, 15). """
    poly_inters = Polygon([(-5, -5), (-5, 15), (15, 15), (15, -5), (-5, -5)])
    gdf = gpd.GeoDataFrame(
        [1], geometry=[poly_inters], crs={"init": "epsg:4326"}
    )
    gdf["attr2"] = ["study area"]
    return gdf


@pytest.fixture
def donut_geom(locs_buff, single_rect_poly_gdf):
    """ Fixture for donut geometry objects. """
    donut = gpd.overlay(
        locs_buff,
        single_rect_poly_gdf,
        how="symmetric_difference",
    )
    return donut


@pytest.fixture
def multi_gdf(donut_geom):
    """ Create a multi-polygon GeoDataFrame. """
    multi_poly = donut_geom.unary_union
    out_df = gpd.GeoDataFrame(
        geometry=gpd.GeoSeries(multi_poly), crs={"init": "epsg:4326"}
    )
    out_df = out_df.rename(columns={0: "geometry"}).set_geometry("geometry")
    out_df["attr"] = ["pool"]
    return out_df


@pytest.fixture
def multi_line(linez_gdf):
    """ Create a multi-line GeoDataFrame.

    This has one multi line and another regular line.
    """
    # Create a single and multi line object
    multiline_feat = linez_gdf.unary_union
    linec = LineString([(2, 1), (3, 1), (4, 1), (5, 2)])
    out_df = gpd.GeoDataFrame(
        geometry=gpd.GeoSeries([multiline_feat, linec]),
        crs={"init": "epsg:4326"},
    )
    out_df = out_df.rename(columns={0: "geometry"}).set_geometry("geometry")
    out_df["attr"] = ["road", "stream"]
    return out_df


@pytest.fixture
def multi_point(locs_gdf):
    """ Create a multi-line GeoDataFrame. """
    multi_point = locs_gdf.unary_union
    out_df = gpd.GeoDataFrame(
        gpd.GeoSeries(
            [multi_point, Point(2, 5), Point(-11, -14), Point(-10, -12)]
        ),
        crs={"init": "epsg:4326"},
    )
    out_df = out_df.rename(columns={0: "geometry"}).set_geometry("geometry")
    out_df["attr"] = ["tree", "another tree", "shrub", "berries"]
    return out_df


def test_not_gdf(single_rect_poly_gdf):
    """Non-GeoDataFrame inputs raise attribute errors."""
    with pytest.raises(AttributeError):
        cl.clip_shp((2, 3), single_rect_poly_gdf)
    with pytest.raises(AttributeError):
        cl.clip_shp(single_rect_poly_gdf, (2, 3))


def test_returns_gdf(locs_gdf, single_rect_poly_gdf):
    out = cl.clip_shp(locs_gdf, single_rect_poly_gdf)
    assert hasattr(out, "geometry")


def test_non_overlapping_geoms():
    """Test that a bounding box returns error if the extents don't overlap"""
    unit_box = Polygon([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)])
    unit_gdf = gpd.GeoDataFrame(
        [1], geometry=[unit_box], crs={"init": "epsg:4326"}
    )
    non_overlapping_gdf = unit_gdf.copy()
    non_overlapping_gdf = non_overlapping_gdf.geometry.apply(
        lambda x: shapely.affinity.translate(x, xoff=20)
    )
        cl.clip_shp(unit_gdf, non_overlapping_gdf)


def test_input_gdfs(single_rect_poly_gdf):
    """Test that function fails if not provided with 2 GDFs."""
    with pytest.raises(AttributeError):
        cl.clip_shp(list(), single_rect_poly_gdf)
    with pytest.raises(AttributeError):
        cl.clip_shp(single_rect_poly_gdf, list())


def test_clip_points(locs_gdf, single_rect_poly_gdf):
    """Test clipping a points GDF with a generic polygon geometry."""
    clip_pts = cl.clip_shp(locs_gdf, single_rect_poly_gdf)
    assert len(clip_pts.geometry) == 3 and clip_pts.geom_type[1] == "Point"


def test_clip_poly(locs_buff, single_rect_poly_gdf):
    """Test clipping a polygon GDF with a generic polygon geometry."""
    clipped_poly = cl.clip_shp(locs_buff, single_rect_poly_gdf)
    assert len(clipped_poly.geometry) == 3
    assert all(clipped_poly.geom_type == "Polygon")


def test_clip_multipoly(multi_gdf, single_rect_poly_gdf):
    """Test a multi poly object can be clipped properly.

    Also the bounds of the object should == the bounds of the clip object
    if they fully overlap (as they do in these fixtures). """
    clip = cl.clip_shp(multi_gdf, single_rect_poly_gdf)
    assert hasattr(clip, "geometry")
    assert np.array_equal(clip.total_bounds, single_rect_poly_gdf.total_bounds)
    # 2 features should be returned with an attribute column
    assert len(clip.attr) == 2


def test_clip_single_multipolygon(locs_buff, smaller_clip_rect_poly_gdf):
    """Test clipping a multi poly with another poly that

    no sliver shapes should be returned in this clip. """

    multi = locs_buff.dissolve(by="type").reset_index()
    clip = cl.clip_shp(multi, smaller_clip_rect_poly_gdf)

    assert hasattr(clip, "geometry") and clip.geom_type[0] == "Polygon"


def test_clip_multiline(multi_line, single_rect_poly_gdf):
    """Test that clipping a multiline feature with a poly returns expected output."""

    clip = cl.clip_shp(multi_line, single_rect_poly_gdf)
    assert hasattr(clip, "geometry") and clip.geom_type[0] == "MultiLineString"


def test_clip_multipoint(single_rect_poly_gdf, multi_point):
    """Clipping a multipoint feature with a polygon works as expected.

    should return a geodataframe with a single multi point feature"""

    clip = cl.clip_shp(multi_point, single_rect_poly_gdf)

    assert hasattr(clip, "geometry") and clip.geom_type[0] == "MultiPoint"
    assert hasattr(clip, "attr")
    # All points should intersect the clip geom
    assert all(clip.intersects(single_rect_poly_gdf.unary_union))


def test_clip_lines(linez_gdf, single_rect_poly_gdf):
    """Test what happens when you give the clip_extent a line GDF."""
    clip_line = cl.clip_shp(linez_gdf, single_rect_poly_gdf)
    assert len(clip_line.geometry) == 2
```

After refactoring that, I am also noticing that the fixture called `smaller_clip_rect_poly_gdf` could have a better name. If I understand correctly its actually **larger** than the other rectangle polygon gdf fixture, instead of smaller as the name might suggest.

earthpy/tests/test_clip.py
mbjoseph: See above for a way to eliminate these helper functions!

earthpy/tests/test_clip.py
lwasser: ok!! i'll be on mission fixture cleanup. i'll have a look at the `smaller_clip_rect_poly_gdf`. it's totally possible that it was opposite day when i created it. i wanted to create a poly that did not test the edge case of a clip of a mult-poly object that returns slivers (lines) and polygons. 

thanks @mbjoseph !!

lwasser: i just pushed a set of changes. i want to give this one more once over because i keep finding issues, duplications and points of clarification

earthpy/tests/test_clip.py
lwasser: done!!

lwasser:
ok @mbjoseph all tests seem to be passing and i think your feedback has made this PR much cleaner. Can you please have another look at it?

earthpy/clip.py
```diff
@@ -146,14 +217,14 @@ def clip_shp(shp, clip_obj):
...
 
-    # Multipolys / point / line don't clip properly
-    if "Multi" in str(clip_obj.geom_type) or "Multi" in str(shp.geom_type):
-        raise ValueError(
-            "Clip doesn't currently support multipart geometries. Consider "
-            "using .explode to create unique features in your GeoDataFrame"
-        )
-
-    if shp["geometry"].iloc[0].type == "Point":
+    # if there are multi points
```
mbjoseph: I think we should delete this comment, because it's redundant with the line below
```suggestion
```

lwasser:
yay thanks so much @mbjoseph  this PR was a good one because i think

1. you helped me clean up the code a LOT. i just hadn't thought that zero row concats would work and yet it did. And the fixtures are all much cleaner now. 
2. it's cool new functionality. 

thank you!! once all tests pass i'll merge and will bump a version up!! 
i just deleted that comment. are you comfortable with approving the PR once tests pass?? (note i see appveyor is not happy so i'll dig into that now).

lwasser:
oh it's just not happy cause there are several builds in the queu for some reason.

lwasser:
ok @mbjoseph  all tests are passing now. Are you comfortable with approving this? then i'll merge and bump :)

mbjoseph:
Looks good @lwasser!

lwasser:
@nkorinek thank you again for your work on this new functionality in earthpy!! we just merged the pr!! hooray!!

Issue #319, opened 2019-05-09
.io module tests failing sometimes


lwasser:
**Describe the bug**
hey @mbjoseph  we are having issues with some of the io tests failing intermittently.  i can restart the build and often it fixes things. it is (i think) related to downloaded things but i'm not sure. 

https://travis-ci.org/earthlab/earthpy/jobs/529901318

could you have a look? the link above is a mac build so it's not the best example. when this happens again i will be sure to copy in the link and tests it's failing on with the associated error.

**To Reproduce**

**Expected behavior**

**Screenshots**

**What Operating System Are you Running? **
 - Mac
 - Windows
 - Linux

**Additional context**

mbjoseph:
@lwasser these tests hit figshare over the internet to check http responses and/or download files. Sometimes connectivity issues cause test failures. Some ideas with pros & cons include:

1. **Skip the tests if no internet connection is detected**. This will not handle cases where figshare's servers are having issues, but might help if most of the errors result from connectivity issues on Travis' end. 
2. **Skip the tests on CI**. This will prevent such failures entirely on Travis CI, but we'll take a coverage hit. 
3. **Commit the response data to the repo** by adding/committing the `earthpy/tests/cassettes/` directory. This will use a "recorded" response, so that the test runs on CI without needing to hit the figshare servers, but will add 536.4 kB worth of files to the repo, and adds more stuff to keep up to date/maintain. 

Any preference for one of these options, or alternative ideas?

mbjoseph:
We decided to go with option 2 - I can look into this and submit a PR.

lwasser:
thanks @mbjoseph  just merged this PR and will close the issue!!

PR #320, opened 2019-05-10
Conditional skip tests on CI

mbjoseph:
Solves #319 by skipping tests with intermittent failures.


mbjoseph:
@lwasser please have a gander at this PR when you get a chance! It conditionally skips the tests requiring internet if running on a CI system.

lwasser:
cool! i see what you did there @mbjoseph another decorator to skip tests. that's good to know we can do if need be. i'll merge this puppy!! thank you!

PR #324, opened 2019-05-20
Add URL to io.py for landsat vignette data

### LINKED ISSUE #309: Make a custom data folder for the vignettes
Right now, we are using `et.data.get_data()` data for the vignettes, and making it work. However, there are sometimes when this is less then ideal, such as having to use `cs-test-landsat` data, which is mostly intended for private use. If we curated some data just for the vignettes we could simplify them and make the data perfect for our examples.

jlpalomino:
@lwasser This pull request adds the url for the landsat vignette data to DATA_URLS in earthpy/io.py.  This data has been successfully tested with the crop data vignette created by @nkorinek. The dataset is approximately 85 MB and covers the Cold Springs fire boundary and includes nodata values at the southeast edge of the image. Fixes #309


jlpalomino:
@lwasser This is PR is now ready for review and merging.

lwasser:
thank you @jlpalomino looks good. i haven't played with the data but it downloaded just fine. let's incorporate this into our vignettes now!! @nkorinek please update vignettes accordingly!! let's see if the docs build.

nkorinek:
@jlpalomino @lwasser Hey, so I updated my earthpy and implemented the new datasource in both vignette 3 and vignette 5. They look fine! The dataset makes it really clean, thanks for the update @jlpalomino ! They should both be ready for final checks before merging.

lwasser:
oh yay @nkorinek the docs build was failing because of that larger data i think... so i'm so glad you think the smaller subset works well because it will help with our build too :) i'll have a look at this hopefully today and if not i'll try to get to it before we meet!

PR #328, opened 2019-05-23
Add plot_rgb vignette

### LINKED ISSUE #306: Vignette #4 -- Plot_rgb vignette
Let's make a vignette on the `plot_rgb()` function that shows
1. how to select different bands to plot band combinations like CIR, RGB
2. how to play with the stretch
3. show how you can use plot bands in a multi panel axis setup

`f, (ax1, ax2) = plt.subplots()...`

we can use @nkorinek stack / crop vignette as a starting place without needing the crop part.

jlpalomino:
@lwasser This pull request adds the plot_rgb vignette and updates the CHANGELOG and the README with this new vignette. Fixes #306


lwasser:
thank you @jlpalomino will you please update from master today? it looks like there are some conflicts to resolve. if you resolve them locally they shouldn't push to master directly also!

examples/plot_rgb.py
```diff
@@ -0,0 +1,168 @@
+"""
+Plot Combinations of Raster Bands Using EarthPy
```
lwasser: This is a good start. i think we should add a section on using the extent parameter. please look at the parameters and make sure the key ones are covered in the vignette!

examples/plot_rgb.py
```diff
@@ -0,0 +1,168 @@
...
+    arr_st,
+    rgb=(4, 3, 2),
+    ax=ax,
+    stretch=True,
+    str_clip=0.5,
+    title="Landsat 8 CIR Image with Stretch Applied",
+)
+plt.show()
+
+###############################################################################
```
lwasser: please look at the build of this. When i build i see a very large white space at the top and bottom of the figure. if you set the figsize() to ensure the height is SMALLER THAN the width it will fix the white space issue. 12,6 seemed to be decent for the size on my machine locally.

Screen Shot 2019-05-24 at 9 21 18 AM

lwasser:
@jlpalomino just a note that i just merged two other pr's so you will have to update one last time when you address my comments. nothing else should have to be merged today so you should be free and clear!!

examples/plot_rgb.py
jlpalomino: Added this using the crop vignette as a starting point.

examples/plot_rgb.py
jlpalomino: Fixed!

jlpalomino:
@lwasser I've added the raster extent to plot the polygon on the composite images and modified the size for the multi-plot figure.  This PR is ready for another review.

lwasser:
awesome.. thank you @jlpalomino i made a few tiny edits and once this rebuilds i'll merge!!

lwasser:
merging this to address https://github.com/earthlab/earthpy/issues/306  once it's built

Issue #358, opened 2019-06-06
Can i remove pandas from dev-requirements?


lwasser:
hey @mbjoseph @joemcglinchy it doesn't seem like we need pandas in the dev-requirements as it should install with geopandas and also it's core to anaconda i think?? any thoughts on this? pybot is encouraging me to reduce dependencies via a dozen PR's! 

https://github.com/earthlab/earthpy/blob/master/dev-requirements.txt

any thoughts on removing it?

mbjoseph:
Yeah let's get rid of it!

lwasser:
oh thank goodness for that!! :)

PR #359, opened 2019-06-06
pandas is a redundant requirement for earthpy

lwasser:
geopandas should install pandas but also pandas is a part of anaconda.

this should address #358


lwasser:
note -- let's see if this breaks RTD build!

PR #394, opened 2019-08-01
Fix hist bug where 1 dim arrs don't work and support masked arrs

lwasser:
## TODO List 

- [ ] Add range to the PR
- [ ] Remove **kwargs for now
- [ ] add tests for range once added 
- [ ] test for `colors` when provided as a string.


earthpy/plot.py
```diff
@@ -401,7 +401,14 @@ def plot_rgb(
...
 
 def hist(
-    arr, colors=["purple"], figsize=(12, 12), cols=2, bins=20, title=None
+    arr,
+    colors=["purple"],
+    figsize=(12, 12),
+    cols=2,
+    bins=20,
+    title=None,
+    xlabel="",
```
lwasser: @nkorinek i've added xlabel and ylabel as arguments. 
we should consider using kwargs in the future and i think i know how to do this

```
`foo(bins, other='keyword', args='go', here=':)')` (edited) 
and then they'll show up in your function as (effectively):
```kwargs = {'other': 'keyword', 'args': 'go', 'here': ':)'}```
```

more in the python slack channel where i was talking with carson

```
would it look something like: `foo({"bins":5}`
```def foo(data, **kwargs):
    '''blah blah some stuff here

    Parameters
    -------------
    kwargs: Keyword arguments to pass to the underlying `hist` function
    ...
    # plot a hist
    f, ax = plt.subplots()
     ax.hist()```
    
# call foo
foo(data, bins=5) (edited) 

carson [16:27]
well, so as you can see above, in the function signature you have `**kwargs`... that tells the function to pull all the args provided to the function into a dict called `kwargs`
so you'd call it 'as if` the args were specified in the function signature (edited) 
`foo(bins, other='keyword', args='go', here=':)')` (edited) 
and then they'll show up in your function as (effectively):
```kwargs = {'other': 'keyword', 'args': 'go', 'here': ':)'}```
(edited)

lawasser [16:29]
ok... so let's pretend bins isn't there - i just updated it
i would call `foo(data, bins=5)`
and it  knows that anything after the first parameter should bw a keyword arg
and it puts it into a dic

carson [16:30]
yes

lawasser [16:30]
ok
getting somewhere in understanding this. in the function itself, how does it know to take bins and place it in the hist call?

carson [16:30]
so then in your function:
```...
ax.hist(bins=kwargs.get('bins', None))
...```

lawasser [16:30]
ok

carson [16:30]
or something like that...

lawasser [16:31]
so... does this mean i need to code for any kwarg that a user might provide?

carson [16:31]
you could also do kwargs.pop if you wanted to remove it and use the kwargs for _another_ function later on down the line
naw, you can just code for things you're expecting, and ignore the rest
just make sure you document what you _will_ deal with

lawasser [16:32]
like it seems like `ax.hist(bins=kwargs["bins"])` might work

carson [16:32]
yup it would

lawasser [16:32]
ok

carson [16:32]
i just provided the 'get' version because if you _don't_ have a bins kwarg, the above will error, the 'get' version won't

lawasser [16:32]
ok... so **kwargs isn't going to let ax.hist() access ANY mpl parameter available to the hist method? just ones i account for

carson [16:32]
it'll just default to the default bins for hist

lawasser [16:33]
oh

carson [16:33]
well, you _can_ do this if you want to pass them _all_ in

lawasser [16:33]
if i don't specify **kwargs in the fun definition, .get will work better?

carson [16:33]
```...
ax.hist(**kwargs)
...```

lawasser [16:33]
OH
oh really?

carson [16:33]
that'll just pass them all through

lawasser [16:33]
that's what i was looking for

carson [16:33]
and then hist can deal with it

lawasser [16:34]
oh that's magical

carson [16:34]
yeh, you just have to do both
yeh its cool

lawasser [16:34]
do both as in call it in the func parameters and the call to `.hist`

carson [16:34]
so you need `**kwargs` in your function signature, and then also in your call down to hist
yes

lawasser [16:34]
OK
ok ok i'm following.

carson [16:34]
nice :slightly_smiling_face:
its super handy
especially as you're building new tools and aren't _quite_ sure what you wanna include

lawasser [16:35]
so one last question about kwargs!! this makes sense for a specific method. now what if we want to accept a few other things that are MPL ax methos like for example x and y lims
that's exactly what's happening. i'm realizig some of the plotting is limited as i've hard coded parameters
individual parameters
so if i wanted to provide say a set of kwargs for mpl ax elements like x and y label
is there a clever way to do that as well?
this might get too complicated
.hist makes sense because it's one method with a set of parameters a user might want to use

carson [16:43]
in that case, if you have _specific_ things you want, you can use the pop thing I mentioned
```...
ylim = kwargs.pop('ylim', None)
xlim = kwargs.pop('xlim', None)
ax.use_lims_somehow(ylim, xlim)
ax.hist(**kwargs)
...```
(edited)

carson [16:44]
_now_ you've pulled out the lim args and used them in ax, but left the rest to be used in hist
is that what you were thinking?

lawasser [16:45]
yea... trying to understand how you might pass variables to different parts of mpl methods
in your example above
what does `None` do?
wouldn't it have a value associated with it?

carson [16:45]
just some useful default value
its just in case they "don't" provide a kwarg of ylim (edited) 

lawasser [16:46]
OH
so would you need a conditional there?
so if they DO provide a val,
if kwargs["xlim"]: or soemthing to test it exists?
i think i'm generally following you however

carson [16:47]
yeh for some things you'd want to do that yes
for some things it probs doesn't matter? or the underlying function knows what to do with None, but yeh, you might do a conditional check
but i like pop because it pulls it _out_ of kwargs, and then you can work with the variable nicely
without the `dict['name']` stuff
but either way obviously works
```

earthpy/plot.py
```diff
@@ -418,9 +425,13 @@ def hist(
         The number of columns for plot grid.
     bins : int (default = 20)
         The number of bins to generate for the histogram.
-    title : list (optional)
+    title : str or list (optional)
```
lwasser: we now support title as a list or string

earthpy/plot.py
```diff
@@ -449,13 +460,25 @@ def hist(
...
+            title = [title]
+
     # If the array is 3 dimensional setup grid plotting
     if arr.ndim > 2:
+        # Compress the arr if it's masked
+        if np.ma.is_masked(arr):
+            arrlis = []
+            for i in range(arr.shape[0]):
+                # Use compressed to flatten masked arr
+                arrlis.append(arr[i].compressed())
```
lwasser: here i used `compressed()` which is a method for masked_arrays. unfortunately you cant specify an axis or else this could be a `try` statement. i'm wondering if there is a way to run compressed on an array rather than looping. i'll ask on slack.

earthpy/plot.py
```diff
@@ -477,17 +500,30 @@ def hist(
...
+            arr_comp = arr.compressed()
+        else:
+            arr_comp = arr.ravel()
+
         fig, ax = plt.subplots(figsize=figsize)
         ax.hist(
-            arr.ravel(),
-            range=[np.nanmin(arr), np.nanmax(arr)],
+            arr_comp,
+            range=[np.nanmin(arr_comp), np.nanmax(arr_comp)],
```
lwasser: range is not an option in this function now but we could add it.

earthpy/tests/test_hist.py
```diff
@@ -0,0 +1,155 @@
...
+    for an_axis in ax:
+        assert len(ax[0].patches) == nbins
+
+    assert len(f.axes) == 2
+    plt.close(f)
+
+
+def test_hist_1band_masked_array(image_array_single_band):
+    """Ensure that a masked single band arr plots & the number of bins is correct"""
+    masked_arr = np.ma.masked_where(
```
lwasser: in your tests @nkorinek essentially you didn't actually mask any values, you just tried to coerce the arr to a masked array. that is not what we needed! here i took an existing fixture and turned it into a masked array by setting values == 4 to masked! then i tested.

earthpy/tests/test_hist.py
```diff
@@ -0,0 +1,155 @@
...
+        image_array_single_band == 4, image_array_single_band
+    )
+    nbins = 3
+    f, ax = ep.hist(masked_arr, title=["TITLE HERE"], bins=nbins)
+    assert len(f.axes) == 1
+    assert len(ax.patches) == nbins
```
lwasser: this step ensures that the number of bars in the plot equals the number of bins that i specified!!

earthpy/tests/test_plot.py
```diff
@@ -24,87 +24,3 @@ def test_colorbar_raises_value_error():
     with pytest.raises(AttributeError, match="requires a matplotlib"):
         ep.colorbar(list())
     plt.close()
-
-
-""" Hist tests """
```
lwasser: i moved all of the hist tests from this file into a `test_hist` file where it should be!

earthpy/tests/test_hist.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_hist_plot_1_dim(image_array_2bands):
+    array_1_dim = image_array_2bands.ravel()
+    f, ax = ep.hist(array_1_dim)
+    assert len(f.axes) == 1
+    plt.close(f)
+
+
+# TODO: test that adding an x and y label to the plot works
```
lwasser: @nkorinek would you like to work on writing two more tests? we should test that the x and y labels actually work. note that i didn't implement this on a multi band plot just a single so far. you are welcome to add it to multiband if you'd like!! 

below we need a test that ensures a color provided as a string works! you may need to add a test to the hist function as well to coerce a color provided as a string to a list just like i did with the title.

lwasser:
there ar ea few things you can work on in this PR @nkorinek !! let's talk tomorrow about what you will work on next given i know you are leaving next week for vacation!!

lwasser:
to fix this code cov gap

https://codecov.io/gh/earthlab/earthpy/pull/393/src/earthpy/plot.py?before=earthpy/plot.py#L500

we need to create a histogram plot with an odd number of plots relative to cols ...

so if the arr has 3 bands, we want 2 cols and that should do it!

earthpy/plot.py
```diff
@@ -401,7 +401,15 @@ def plot_rgb(
...
-    arr, colors=["purple"], figsize=(12, 12), cols=2, bins=20, title=None
+    arr,
+    colors=["purple"],
+    figsize=(12, 12),
+    cols=2,
+    bins=20,
+    title=None,
+    xlabel="",
+    ylabel="",
+    **kwargs,
```
lwasser: i've added this but it's untested. we should test it using *range* if it works, then we can actually remove bins as an explicit parameter and just include it via kwargs

nkorinek:
## TODO List 
* [ ]  Add range to the PR 
* [ ]  Remove **kwargs for now 
* [ ]  add tests for range once added
* [ ]  test for `colors` when provided as a string.

Issue #390, opened 2019-07-25
Issue with histogram function dealing with 1 dimensional arrays


nkorinek:
Hey! 

I was attempting to use `ep.hist()` to plot a 1 dimensional array and I found an issue with the code. When I was trying to plot the array, the code would run without error, but not histogram would pop up. I couldn't figure out what the issue was, but upon expecting the `ep.hist()` code, I noticed that it never checks that the array has at least two dimensions. There is an `if /elif` statement  in the code. The `if` statement checks to see if there are more than two dimensions, and the `elif` statement checks if there are exactly two dimensions, but no statements actually check that there are more than 2 dimensions. I realize remotely sensed data should always have at least two dimensions, but there should at least be an error message thrown if a 1 dimension array is passed in. Thanks! @lwasser how should we go about fixing this issue?

lwasser:
* install earthpy into the ea-python envt interactively `pip install -e .`
* update the code in earthpy / 
* go back to the notebook, restart the kernel, and rerun cells

- [ ] Handle single dim arrays  if `arr.ndim <= 2` :
   ravel() if arr.ndim==1
- [ ] handle masked arrays- `if ma.is_masked(rgb_bands):`

arr == (masked array)
`arr[~arr.mask]`
*******
- [ ] some parameters are not available 

* hist is also missing several needed kwargs (key word args)

https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.hist.html

nkorinek:
The first half is in this pr right now: https://github.com/earthlab/earthpy/pull/391 . It still needs tests written for it.

lwasser:
ok great! i have more info on the `**kwargs` item but we should discuss it further!

lwasser:
this has been implemented!! closign

PR #402, opened 2019-08-12
Update hist function to support masked arrays and single dim arrays

lwasser:
this addresses #390 

in this PR i've also added support for 3 dimensional arrays that are actually just one band using `np.squeeze()`


lwasser:
This all seems to be working now!! merging and will bump version

PR #422, opened 2019-09-13
fix appveyor fail by installing earthpy without the deps

lwasser:
Essentially this happened for matplotcheck as well. Because earthpy is installing locally via python, it tries to reinstall all of the dependencies. this causes issues. but all of those deps are in our envt so it's unneccesary. installing via pip without the dependencies should make appveyor happy.


lwasser:
@nkorinek i had to fix a small CI bug. can you please update your branches / pr's ?

PR #430, opened 2019-10-19
import earthpy as ep, not et

sgillies:
To fit with the code in the following block, yes?


lwasser:
ahhhh great catch @sgillies  i added `earthpy.plot as ep` as i thijnk that you need to call out the module specifically to access `ep.plot_bands()` or that is how we've been doing it!! We appreciate this catch!!

PR #431, opened 2019-10-19
Add requests to setup requirements

sgillies:
It's imported by earthpy.io but not specified in the setup script, so `pip install earthpy` in a new fresh python environment doesn't quite work. This PR takes care of that.


lwasser:
@sgillies  thank you so much for this PR!! i actually hadn't tested installing earthpy in a clean envt!! merging this!

Issue #433, opened 2019-10-24
Explain relationship that earthpy has or doesn't have to commercial software?


sgillies:
Earthpy explains its relationship to lower-level open source packages like rasterio and geopandas well: it builds on top of them to make some common tasks easier and gives its user more leverage. In reviewing earthpy for https://github.com/pyOpenSci/software-review/issues/3 I wonder if earthpy needs or doesn't need to explain how it relates to commercial software widely used in the higher ed research and teaching domain? This question is not loaded, I promise! 

@lwasser if you don't think this needs to be added to the README, just say so, you are much more of an expert in the teaching field, and I will tick the box in the review.

lwasser:
hi @sgillies !! this is a great question and i'd love to better understand what you are thinking!

What types of commercial software are you thinking about? Like GIS type GUI based tools ??

We modeled earthpy after the `raster` and `sp` packages in `R`. These tools made is a lot easier to handle spatial manipulations in the R language that that didn't' seem to exist for python! 

As far are commercial software that may be related - i am not sure what tools I would call attention to but i'm very open to suggestions if you had any in mind!! i'm not adverse to updating the readme if it would make the goal of our package more clear. @joemcglinchy @mbjoseph @jlpalomino  can you have a look at this issue and see what you think about other packages that might put earthpy into better context? The R tools are the ones that come to mind for me.

sgillies:
@lwasser I think it would be very good if the R packages this is modeled after were mentioned, and if I read the intent of the package correctly, that earthpy is intended for interactive use in a notebook. If earthpy wasn't meant to replace or integrate with commercial software, then no need to mention any.

mbjoseph:
Hi @sgillies and @lwasser! Some thoughts here: 

- I agree we should mention the R packages that inspired EarthPy. Both `raster` and `sp` are wrappers around lower level libraries, but provide a higher level of abstraction, which makes them a bit less flexible but perhaps easier to use for beginners. EarthPy is similar in this sense.
- We tend to use EarthPy in notebooks (in the classroom), but it will work in scripts as well. We don't (AFAIK) have any functionality that is directly tied to a notebook execution environment.
- EarthPy isn't intended to integrate with or replace any particular commercial GIS software in general, so I'd lean towards not mentioning it.

Thanks @sgillies for a thoughtful/comprehensive review! We really appreciate it.

lwasser:
[...]

Issue #432, opened 2019-10-19
Because of rtree, earthpy pip installs on OS X aren't complete


sgillies:
**Describe the bug**

Earthpy depends on rtree, which requires libspatialindex, which is not provided by the wheels on PyPI. Until the rtree wheels do include their own requirements, it's on projects like earthpy to explain the dependency.

lwasser:
ahhhhh @sgillies gosh i didn't realize this was an issue. it must be that libspatialindex was installing for us automagically in the dev environment. 

Can you help me understand how you'd like to see this dependency explained? 

i see in the rtree readme

https://pypi.org/project/Rtree/

it just calls it out as a dependency. is that the type of thing that you are looking for here?

Again many thanks for this input!

sgillies:
Yes, I think that would be a good start. But because earthpy is intended to be higher level, for users who aren't necessarily C programmers or sysadmins (like rtree's original audience), some additional documentation may be good. 

Someone who is new to Python and doesn't know what libspatialindex is or where they can get it is in for a hard time when they try to install earthpy from PyPI. Could you make this a conda only package? Is that realistic?

lwasser:
@sgillies do you mean just encourage installation from conda rather than pip? i think that conda is the best way to install given all of the other spatial dependency issues ! you are correct in that we assume our audience are NOT c programmers. we use this package with students who are quite new to programming in general!! 

so documentation could say something like.. we suggest that you install earthpy using conda (conda-forge is our preferred channel) but then given it is on pypi, i could add some language about libspatialindex with a "proceed with caution" stipulation or something?

would that be a reasonable approach?

thank you AGAIN.

sgillies:
@lwasser totally reasonable 👍

lwasser:
ok closing this as well as it was addressed in #440. please reopen if you'd like to see more explanation @sgillies !!

PR #440, opened 2019-10-25
Update readme with conda vs pip instructions

lwasser:
This PR addresses #432 created by @sgillies . Earthpy has a dep on the rtree library which requires libspatialindex. this dependency does not install via pip and thus could cause issues for users. 

In the readme, I've added instructions which tell the user to install via conda if they can. this is the use case for most of our students given they use conda so it seems reasonable. 

I've also added a statement associated with earthpy being inspired by the R raster and sp packages which should address #433 

@mbjoseph can you have a look at this PR when you have a chance? @sgillies please let me know if these changes are what you had in mind or if you'd like to see a bit more!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,9 @@ of our contributors.
 ## Contributors
 
 We've welcome any and all contributions. Below are some of the
-contributors to EarthPy.
+contributors to EarthPy. We are currently trying to update this list!!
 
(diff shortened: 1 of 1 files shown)


lwasser:
thank you @mbjoseph merging now... but will see if @sgillies would like any additional readme changes to satisfy those two issues opened!! Happy to add more!

Issue #434, opened 2019-10-24
Within et.data.get_data(), consider using Python's logging instead of print()


sgillies:
Context: https://github.com/pyOpenSci/software-review/issues/3

When get_data() is called, text is printed to stdout.

```
>>> data_path = et.data.get_data("vignette-landsat")
Downloading from https://ndownloader.figshare.com/files/15197339
Extracted output to /Users/seang/earth-analytics/data/vignette-landsat/.
```

On the one hand, some feedback to users is a good thing. On the other hand, library functions should avoid printing to stdout and should use Python's logging facility instead so that users have more control over the output.

sgillies:
That said, I scanned the package and this is the only use of print that I could find. If earthpy is mostly for interactive active use in a notebook and not for developing non-interactive programs, calling print from get_data is harmless. Feel free to close this one.

lwasser:
hey @sgillies yes we mostly intend get_data to be used in an interactive environment.  so perhaps we shouldn't modify this now but note that in the future if users want to use it more in a non interactive environment, we may want to get rid of the print statements? 

the other option is to set `verbose=False` for non interactive applications? is that bad practice?

sgillies:
The option is good, but logging (where the default stream is stderr) would be even better. The situation that we want to prevent is where someone writes a script with earthpy that prints data to stdout (used like `python myscript.py > data.geojson`) and it has data download messages interleaved with the JSON text output.

lwasser:
thank you again, @sgillies ok i did some reading on this and started a PR 

https://github.com/earthlab/earthpy/pull/448 

does this approach seem reasonable? OR what would you implement here? in this case i printed if `verbose=True` (which is a default) but used `logging.info` if it's false. Please know this particular concept is new to me SO if i am implementing it incorrectly please do say the word or suggest another approach!!!

sgillies:
@nkorinek yes, changing the log level within the function is a good approach. Instead of configuring the root logger, I think the project should set the level of the module's logger `logging.getLogger(__name__).setLevel(logging.INFO)`. This will increase earthpy's verbosity only and not add noise from requests or other modules.

nkorinek:
@sgillies that looks much cleaner, cool!

lwasser:
ok thank you again for this suggestion @sgillies  for the time being i'm going to close this and the associated PR. but perhaps we will revisit it in the future if it becomes a real problem for people using the package!! your input and review has been greatly appreciated!! THANK YOU!

PR #448, opened 2019-10-28
add logging if verbose is set to false

lwasser:
this addresses #434 
the idea being that if the user turns off verbose, logging will be implemented instead to make it easier to troubleshoot via scripts.

### DIFF
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -320,7 +320,8 @@ class Data(object):
         if verbose is True:
             print("Extracted output to {}".format(path))
         else:
-            loggin.info("Extracted output to {}".format(path))
+            logger = logging.getLogger(name)
+            logger.info("Extracted output to {}".format(path))
 
 
 def path_to_example(dataset):
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -94,7 +94,7 @@ DATA_NAME = op.join("earth-analytics", "data")
 ALLOWED_FILE_TYPES = ["file", "tar", "tar.gz", "zip"]
 
 # Create logger object
-logger = logging.getLogger(name)
+logger = logging.getLogger("log_name")
 
 
 class Data(object):
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -275,7 +275,6 @@ class Data(object):
 
         if verbose is True:
             print("Downloading from {}".format(url))
-        else:
             logger.info("Downloading from {}".format(url))
 
         r = requests.get(url)
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -274,7 +274,7 @@ class Data(object):
             return path
 
         if verbose is True:
-            print("Downloading from {}".format(url))
+            logger.setLevel(logging.INFO)
         logger.info("Downloading from %r", url)
 
         r = requests.get(url)
--- a/earthpy/io.py
+++ b/earthpy/io.py
@@ -95,6 +95,7 @@ ALLOWED_FILE_TYPES = ["file", "tar", "tar.gz", "zip"]
 
 # Create logger object
 logger = logging.getLogger(__name__)
+original_level = logger.level
 
 
 class Data(object):
@@ -183,7 +184,10 @@ class Data(object):
         if key is None and url is None:
             # If user doesn't input anything, print & log all available data keys.
             print(self.__repr__())
-            logger.info(self.__repr__())
+            # with LoggingContext(logger, logging.INFO if verbose else None):
+            #    logger.info("message")
+            # logger.setLevel(logging.INFO)
+            # logger.info(self.__repr__())
             return
 
         if key is not None:
@@ -205,7 +209,7 @@ class Data(object):
                 else:
                     fname = url.split("/")[-1]
 
-            # try and deduce filetype based on extension
+            # Try and deduce filetype based on extension
             file_type = "file"
             for ext in ALLOWED_FILE_TYPES:
                 if fname.endswith(ext):
@@ -287,7 +291,7 @@ class Data(object):
             self._download_and_extract(path, r, kind, verbose)
         return path
 
-    def _download_and_extract(self, path, r, kind, verbose):
+    def _download_and_extract(self, path, r, kind, verbose=True):
         """ Download and extract a compressed archive.
 
         This function downloads and extracts compressed directories to
@@ -319,7 +323,11 @@ class Data(object):
             archive = tarfile.open(fileobj=file_like_object, mode="r:gz")
         os.makedirs(path, exist_ok=True)
         archive.extractall(path)
-        # Logger is set to info level above so no need to set it again
+        # We want the logger to PRINT if verbose is set to true, and just log if it's false.
+        # this does NOT seem to be printing
+        if verbose:
+            logger.setLevel(INFO)
+
         logger.info("Extracted output to {}".format(path))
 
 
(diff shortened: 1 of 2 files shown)


earthpy/io.py
```diff
@@ -313,6 +319,8 @@ def _download_and_extract(self, path, r, kind, verbose):
         archive.extractall(path)
         if verbose is True:
             print("Extracted output to {}".format(path))
+        else:
+            loggin.info("Extracted output to {}".format(path))
```
kcranston: typo here (`logging`, not `loggin`)

lwasser: fixed!!

lwasser:
ahhh i think code cov is unhappy because i've added a bit more code with no associated tests... will revisit this tomorrow!

earthpy/io.py
```diff
@@ -91,6 +93,9 @@
 DATA_NAME = op.join("earth-analytics", "data")
 ALLOWED_FILE_TYPES = ["file", "tar", "tar.gz", "zip"]
 
+# Create logger object
+logger = logging.getLogger("io")
```
sgillies:
```suggestion
logger = logging.getLogger(__name__)
```

Best practice is to use `__name__`.

lwasser: thank you @sgillies !!! accepted this suggestion

earthpy/io.py
```diff
@@ -268,6 +275,8 @@ def _download(self, url, path, kind, replace, verbose):
 
         if verbose is True:
             print("Downloading from {}".format(url))
```
sgillies:
```suggestion
            logger.setLevel(logging.INFO)
```

earthpy/io.py
```diff
@@ -268,6 +275,8 @@ def _download(self, url, path, kind, replace, verbose):
...
+        else:
+            logger.info("Downloading from {}".format(url))
```
sgillies:
```suggestion
        logger.info("Downloading from %r", url)
```

earthpy/io.py
```diff
@@ -268,6 +275,8 @@ def _download(self, url, path, kind, replace, verbose):
 
         if verbose is True:
             print("Downloading from {}".format(url))
+        else:
```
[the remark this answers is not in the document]
lwasser: ok so here @sgillies if i understand correctly, we are removing the conditional, and it will log regardless which is good. but what allows it to print a message out for the user? bare with me - this logging approach is new and i'm learning here a lot!!

lwasser:
thank you @sgillies 
is it better to set the logging level here: 

https://github.com/earthlab/earthpy/pull/448/files#diff-de9e7dc8021e4b699c24f60824d933a0R277 

```python
        if verbose is True:
            logger.setLevel(logging.INFO)
        logger.info("Downloading from %r", url)
```

vs at the top where the logger is defined?
https://github.com/earthlab/earthpy/pull/448/files#diff-de9e7dc8021e4b699c24f60824d933a0R97
```
logger = logging.getLogger(__name__)
# could be:??
logger = logging.getLogger(__name__).setLevel(logging.INFO)
```

i'm just trying to understand this a bit better!! 
i really appreciate these suggestions!

sgillies:
@lwasser setting the level of the module's logger to INFO only within the function and `if verbose`, and then setting it back to the original level before the function returns, best preserves the original intent as I understand it. If you set the level on module load, it's too soon.

Here's a barebones example of toggling the log level in a python interpreter.

```
>>> import logging
>>> logging.basicConfig()  # the top level program, for example jupyter notebook, will do this
>>> logger = logging.getLogger(__name__)
>>> logger.info("message")  # no output
>>> original_level = logger.level
>>> logger.setLevel(logging.INFO)
>>> logger.info("message")
INFO:__main__:message
>>> logger.setLevel(original_level)
>>> logger.info("message")  # no output, again
>>>
```

There's a useful example of a context manager that can do this for you in https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging. The "handler" and "close" parts aren't needed and could be deleted. Usage in earthpy would be something like:

```
with LoggingContext(logger, logging.INFO if verbose else None):
    log.info("message")
```

nkorinek:
@sgillies alright, good to know. Thanks!

lwasser:
got it!! thank you @sgillies !!! i think i understand. the context manager could work well for this use case too!! and thank you again @nkorinek for your work on this pr as well!!

earthpy/io.py
```diff
@@ -91,6 +93,9 @@
 DATA_NAME = op.join("earth-analytics", "data")
 ALLOWED_FILE_TYPES = ["file", "tar", "tar.gz", "zip"]
 
+# Create logger object
+logger = logging.getLogger(__name__)
```
lwasser: here we create the logger -- this works but this logger is at the default LEVEL so it does not print.

earthpy/io.py
```diff
@@ -267,7 +274,8 @@ def _download(self, url, path, kind, replace, verbose):
             return path
 
         if verbose is True:
```
lwasser: should we try: 
```
with LoggingContext(logger, logging.INFO if verbose else None):
    log.info("message")
```

lwasser: this is NOT printing is the issue right now. we want the logger to PRINT the output IF verbose is set to TRUE

lwasser: This doesn't print the log statement regardless of how we configure things. it's almost as if setting the level is not working.

earthpy/io.py
```diff
@@ -265,9 +279,12 @@ def _download(self, url, path, kind, replace, verbose):
         path = op.expanduser(path)
         if replace is False and op.exists(path):
             return path
-
-        if verbose is True:
-            print("Downloading from {}".format(url))
+        # We know here the level is being set to INFO but it is still not printing (as a log) in the terminal
+        if verbose:
```
lwasser: @sgillies if you have time and patience we are struggling a bit with logging. In this example, we are printing out the level to ensure that the level is "INFO". however even with it being set to INFO, it does NOT seem to print the log out for the user. do you happen to know why this might be happening? it DOES work if we set the logger to INFO at the root using `logging.basicConfig()` 

Any thoughts on this?

lwasser:
@sgillies our second question for you deals with tests. What is the best way to test for log output using pytest. We have tried using the `caplog` and  `capsys` fixtures to do this however neither seem to actually capture the output. Neither of these fixtures test whether the log output is printed in the terminal for the user to see. We just want to ensure that we have proper test coverage. 

Many thanks for your patience in helping us here! We also think that perhaps the other question that I asked about the logger not printing for the user COULD help us with this test as well but we are not sure!

earthpy/tests/test_io.py
```diff
@@ -220,6 +220,15 @@ def test_url_download_txt_file_with_content_disposition(eld):
     assert path.endswith("abc.txt") and os.path.isfile(path)
 
 
+@skip_on_ci
+@pytest.mark.vcr()
```
lwasser: This test captures the log itself. but what if we want to capture that the log message prints to the console if verbose=True?

sgillies:
@lwasser I regret suggesting the use of python's logger and think you should go back to calling print(). Sorry about the distraction.

lwasser:
ok thank you @sgillies !!

lwasser:
closing this because of several technical issues. we may revisit in the future!

Issue #507, opened 2020-03-22
earthpy not installed via pip


rlzijdeman:
**Describe the bug**
Using `pip install earthpy` or `pip install earthpy --upgrade` I get the following error:
```
xxx@xxx-mbp-2 gisPython % pip install earthpy --upgrade
Collecting earthpy
  Using cached earthpy-0.8.0.tar.gz (1.4 MB)
Requirement already satisfied, skipping upgrade: geopandas in /Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages (from earthpy) (0.7.0)
Requirement already satisfied, skipping upgrade: matplotlib>=2.0.0 in /Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages (from earthpy) (3.2.1)
Requirement already satisfied, skipping upgrade: numpy>=1.14.0 in /Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages (from earthpy) (1.18.2)
Requirement already satisfied, skipping upgrade: rasterio in /Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages (from earthpy) (1.1.3)
Collecting Rtree>=0.8
  Using cached Rtree-0.9.4.tar.gz (62 kB)
    ERROR: Command errored out with exit status 1:
     command: /Library/Frameworks/Python.framework/Versions/3.8/bin/python3.8 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/zr/vrpsbc0104v_k5nx4dnhpt580000gn/T/pip-install-p57jsw00/Rtree/setup.py'"'"'; __file__='"'"'/private/var/folders/zr/vrpsbc0104v_k5nx4dnhpt580000gn/T/pip-install-p57jsw00/Rtree/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/zr/vrpsbc0104v_k5nx4dnhpt580000gn/T/pip-install-p57jsw00/Rtree/pip-egg-info
         cwd: /private/var/folders/zr/vrpsbc0104v_k5nx4dnhpt580000gn/T/pip-install-p57jsw00/Rtree/
    Complete output (15 lines):
    Traceback (most recent call last):
[...]

PR #509, opened 2020-03-24
Deprecate Clip: remove rtree which is causing pip install issues

lwasser:
i'm not sure we are even still using it so i'm going to test a local remove to see what happens.

this relates to #507 and should potentially fix it but we need to see how the docs, etc build as well


lwasser:
failing because rtree is missing. and likely it can't install properly via pip as seen locally.
Because we are deprecating clip, we may be able to circle around this however because the clip tests are what is failing.

lwasser:
ok. CI is failing here because codecov is simply not working. But this PR actually REMOVES an entire module and so cover coverage is not relevant for this PR. All testing is passing so i'll merge this even tho it looks "broken" it is not. ci is struggling today. as is my internet.

Issue #512, opened 2020-03-25
Earthpy Get_data key for twitter data is incorrect


lwasser:
**Describe the bug**
If you try to download the twitter-flood data, get_data fails returning:
To me, it looks like it's trying to unzip a file that is not a zip file.

```
Downloading from https://ndownloader.figshare.com/files/10960175

---------------------------------------------------------------------------
BadZipFile                                Traceback (most recent call last)
<ipython-input-36-45a13e0cd8e9> in <module>
      1 import earthpy as et
----> 2 et.data.get_data("twitter-flood")

~/miniconda3/envs/earth-analytics-python/lib/python3.7/site-packages/earthpy/io.py in get_data(self, key, url, replace, verbose)
    229                 kind=kind,
    230                 replace=replace,
--> 231                 verbose=verbose,
    232             )
    233             data_paths.append(this_path)

~/miniconda3/envs/earth-analytics-python/lib/python3.7/site-packages/earthpy/io.py in _download(self, url, path, kind, replace, verbose)
    277                 f.write(r.content)
    278         else:
--> 279             self._download_and_extract(path, r, kind, verbose)
    280         return path
    281 

~/miniconda3/envs/earth-analytics-python/lib/python3.7/site-packages/earthpy/io.py in _download_and_extract(self, path, r, kind, verbose)
    305         file_like_object = io.BytesIO(r.content)
    306         if kind == "zip":
--> 307             archive = zipfile.ZipFile(file_like_object)
    308         if kind == "tar":
    309             archive = tarfile.open(fileobj=file_like_object)

~/miniconda3/envs/earth-analytics-python/lib/python3.7/zipfile.py in __init__(self, file, mode, compression, allowZip64, compresslevel)
   1256         try:
   1257             if mode == 'r':
-> 1258                 self._RealGetContents()
   1259             elif mode in ('w', 'x'):
   1260                 # set the modified flag so central directory gets written

~/miniconda3/envs/earth-analytics-python/lib/python3.7/zipfile.py in _RealGetContents(self)
   1323             raise BadZipFile("File is not a zip file")
   1324         if not endrec:
-> 1325             raise BadZipFile("File is not a zip file")
   1326         if self.debug > 1:
   1327             print(endrec)

BadZipFile: File is not a zip file
```

**To Reproduce**
```python
import earthpy as et
et.data.get_data("twitter-flood")
```
This data should just download into the twitter-flood folder. But instead it can't seem to download.

lwasser:
[...]

PR #513, opened 2020-03-25
BUG FIX: fix twitter-flood data key

lwasser:
This pr addresses #512 which specifically relates to twitter data. since the data are not zipped they needed to be entered into earthpy differently to ensure they were downloaded as files.


nkorinek:
@lwasser so I tested this locally and it seems like your fix works! Sorry I didn't get to it before you did, but I think this issue is good to go!

lwasser:
thank you for the testing, @nkorinek !!! merging this

PR #534, opened 2020-04-27
Earthpy changed to Matplotcheck

beammieb:
@earth-lab This was done as part of my learning on earthdatascience.org.

Thank you.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -2,9 +2,9 @@
 Contributing Guidelines
 =======================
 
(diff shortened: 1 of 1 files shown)


lwasser:
hi @beammieb i think we need to add some clarification to our lessons. you should NEVER submit a pull request like this to a working project. if you want to practice - practice with a colleagues repo who you know. Alternatively you can create a second repo yourself and practice. this PR would actually be BAD if it were merged into earthpy. Also it is not a good use of our time maintaining the package. Closing this.

PR #631, opened 2020-12-02
Update CI to use GitHub Actions

### LINKED ISSUE #630: Switch CI to github actions
currently we are building on travis CI but we need to move away from them given credits cost $$ and we run out given the size of our org.

I have things setup in [abc-classroom](https://github.com/earthlab/abc-classroom/tree/main/.github/workflows) however that is using a pip install and we really should be running conda forge. 

TODOs:

- [x] setup actions to run on conda / miniforge
- [ ] use the same action on abc-classroom to release to pypi -- <will need to add a secret to this repo for that push to pypi on a release>

lwasser:
closes #630 

This updates CI to use actions rather than travis

a few  notes

1. https://github.com/s-weigand/setup-conda/issues/6 recommends using `source activate` given `conda init` requires a bash restart which is tricky. The alternative would be the other conda action which using a $conda envt and path to conda for each call - that seems clunky so going with `source activate`
2. the install is WAY slower with conda vs using pip (look at abc-classroom actions) but i think we need to given the many conda dependencies. 

We don't need to lint and build docs on each version so i pulled them out into a separate build. this simplifies the conda environment and that part of the build will be speedy! seems like a win all around. We could consider adding code cov to that build in the future.


lwasser:
this is failing because of appveyor. i've gone ahead and turned off appveyor completely so it wont fail for future builds! all other tests and runs are passing. merging this.

PR #878, opened 2023-06-09
Dependancy Update and Test Fixes

nkorinek:
Implemented developers dependency updates and updated our tests to fix bugs that were showing up when running pytest


nkorinek:
Updated tests and python versions to be more current, updated all dependencies for the earthpy-dev environment

PR #891, opened 2023-07-13
Custom file name for file download

nkorinek:
Added the argument `file_name` that can be used with custom url downloads in `earthpy.data.get_data()`. Allows users to change the name of the final file downloaded. Can't be used with `key` downloads.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`
 unreleased
 ----------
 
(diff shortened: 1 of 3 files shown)


nkorinek:
@eculler if you can check this out and see if it does what you expect it too that would be appreciated!

eculler:
This looks good to me! I will try running it with one of our NCEI downloads, too.

nkorinek:
@eculler let me know the results of that!

eculler (changes requested):
This feature isn't working quite right - it still tacks on http parameters to the file name. See my comment on the Conversation thread.

Issue #905, opened 2023-09-01
Support APPEEARS downloads


eculler:
Earthpy users need to be able to access a selection of NASA remote sensing data in just one step or two. The [APPEEARS API](https://appeears.earthdatacloud.nasa.gov/) is much easier to use than other NASA APIs, and so is a good opportunity to include this feature in earthpy.

@nkorinek

PR #906, opened 2023-09-01
Apppears

eculler:
# Addressing #905 Support APPEEARS Downloads

## Description

This PR provides an AppeearsDownloader class. Users can instantiate the class with information about the dataset, time period, and spatial extent. The Downloader can then be triggered to download (or continue an incomplete download).

The AppeearsDownloader does not expose every feature of the API, but provides enough to be usable. Notably, spatial extent must be specified by GeoDataFrame, not bounding box or point location coordinates.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation update

## How Has This Been Tested?

- [ ] A pytest test has been written, which downloads some Landsat data. This requires the use of the system keyring, and so this test will not run on GitHub Actions yet, but it can be run elsewhere. To test caching, make sure to stop the download before it is complete and re-start.

## Checklist:

- [X] I have already submitted an issue and it was approved for a pr by an earthpy maintainer
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules
- [X] I have checked my code and corrected any misspellings


eculler:
@nkorinek So, I used the `keyring` library to store Earthdata passwords, but honestly this is causing some problems since I can't figure out what the current version of the gnome keyring even is, much less get it installed on Codesaces. So I'm wondering if we can at least give the option to use a plain text configuration file. We could set it up to populate from GitHub secrets for Codespace users? If there's no config file, users will get prompted by `input()`, which is the current no-keyring behavior (but seems to get tripped up by ad blockers and popup blockers)

